PR #12, opened 2022-03-15
Begin fleshing out DNS records

dglsparsons:
Still to do: 
- [x] Validate the schema across fields (i.e. no mx_priority on non-mx records, no srv on non-srv records, no value on srv records). 
- [x] Flesh out the Update method 
- [x] Expand the test suite 
- [x] Add import method for DNS Records
- [x] Add documentation and examples
- [x] Add testing domain env variable to CI

Fixes INT-309


client/dns_record_list.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"strings"
+)
+
+func (c *Client) ListDNSRecords(ctx context.Context, domain, teamID string) (r []DNSRecord, err error) {
+	url := fmt.Sprintf("%s/v4/domains/%s/records?limit=100", c.baseURL, domain)
```
ecklf, line 11: Could this be problematic if someone uses more than 100 domains?

dglsparsons: Good shout, it's currently only used by the sweeper (script to remove any orphaned test resources). So it shouldn't be an issue for anything.

dglsparsons: I'll add a comment explaining this though.

ecklf (approved):

PR #16, opened 2022-04-19
Remove file path trimming for root_directory settings

### LINKED ISSUE #14: Can't use "../" as root_directory
Hey!

It looks like it is not possible to use `"../"` as the project root directory which contradicts with the documentation at the [provider's docs](https://registry.terraform.io/providers/vercel/vercel/latest/docs/resources/project). 

image

However, using an additional path after the upward path, e.g. `"../foo"` seems to work.

```
data "vercel_project_directory" "foo" {
  path = "../"
}

resource "vercel_deployment" "foo" {
  project_id = vercel_project.foo.id
  files      = data.vercel_project_directory.foo.files
  production = true
  project_settings = {
    root_directory = "../"
  }
}
```

```
Could not create deployment, unexpected error: invalid_root_directory - If defined, the Root
Directory must be a relative path not starting with `./` and not including `../` or other special
characters.
```

I often hold my terraform files in a separate folder in my repository while keeping the app at root so this is a bit problematic as I would need to change folder structure. I can work with it for now but I assume this is a bug?

dglsparsons:
- Add a `path_prefix` attribute to a deployment to indicate what fields
  should be trimmed.
- Update documentation indicating the correct use of a
  project_directory.

See https://github.com/vercel/terraform-provider-vercel/issues/14#issuecomment-1103973603 for a full rationale and explanation.

Closes #14


vercel/file_path.go
```diff
@@ -6,13 +6,13 @@ import "strings"
 // This is useful as Vercel doesn't allow a root directory that navigates upwards.
 // But if we trim all file paths upwards, and we also trim the root directory, then
 // all is good.
-func trimFilePath(path string) string {
+func trimFilePath(path string) *string {
 	for strings.HasPrefix(path, "../") {
 		path = strings.TrimPrefix(path, "../")
 	}
 	if path == "" {
-		return "."
```
dglsparsons, line 14: I thought (incorrectly) that we allowed `.` as a root_directory. It turns out the correct thing to do is not to specify it as `.`, but to pass `nil`.

PR #20, opened 2022-04-21
Support per-branch environment variables

### LINKED ISSUE #13: Support per-branch environment variables
We have a use case were we need per-branch variables (A "staging" environment using a different credentials then preview)
https://vercel.com/changelog/environments-variables-per-git-branch

BarnabyShearer:
Fixes #13

Straightforward change to expose the API field.

### DIFF
--- a/docs/resources/project.md
+++ b/docs/resources/project.md
@@ -93,6 +93,7 @@ resource "vercel_project" "example" {
 
 Optional:
 
+- **git_branch** (String) The git branch of the environment variable.
 - **id** (String) The ID of the environment variable
 - **key** (String) The name of the environment variable.
 - **target** (Set of String) The environments that the environment variable should be present on. Valid targets are either `production`, `preview`, or `development`.
--- a/vercel/resource_project.go
+++ b/vercel/resource_project.go
@@ -73,6 +73,11 @@ terraform to your Deployment.
 						},
 						Required: true,
 					},
+					"git_branch": {
+						Description: "The git branch of the environment variable.",
+						Type:        types.StringType,
+						Optional:    true,
+					},
 					"key": {
 						Description: "The name of the environment variable.",
 						Type:        types.StringType,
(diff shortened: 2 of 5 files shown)


docs/resources/project.md
```diff
@@ -93,6 +93,7 @@ resource "vercel_project" "example" {
 
 Optional:
 
+- **git_branch** (String) The git branch of the environment variable.
```
dglsparsons, line 96: I'm not 100% familiar with this API, but can you only specify `git_branch` for `preview` `target`s, or does it work across all different target possibilities?
e.g. what happens if I specified
```
target     = ["production", "preview"]
git_branch = "staging" 
key        = "foo"
value      = "bar" 
```

I'm wondering if it's worth adding some additional validation around that. 
Granted the provider is currently pretty light on up-front validation.

vercel/resource_project.go
```diff
@@ -73,6 +73,11 @@ terraform to your Deployment.
 						},
 						Required: true,
 					},
+					"git_branch": {
+						Description: "The git branch of the environment variable.",
+						Type:        types.StringType,
+						Optional:    true,
+					},
```
dglsparsons, line 80: I think we're also missing a test that includes the `git_branch` being used.

docs/resources/project.md
BarnabyShearer: The API is very strict; the branch has to actually exist, can't be your production branch etc. Lots of things we cannot validate here. You get good error messages if you try to do something impossible.

vercel/resource_project.go
BarnabyShearer: I think at the moment the git tests only actually work for you. I get ```
Could not create project, unexpected error: bad_request - To link a GitHub
        repository, you need to install the GitHub integration first.
```
Which I think is actually because I can only use my own repositories. However I have added some reasonable tests that will hopefully work.

dglsparsons: Ahh yeah, makes sense. We should probably read an environment variable for the non-team git repository or something. I don't think we can keep it pointing at my git repos forever 🤣. One for later though!

dglsparsons (approved):

PR #21, opened 2022-04-22
GitSource: added git source deployment

WCMinor:
Followed the API documentation https://vercel.com/docs/rest-api#endpoints/deployments/create-a-new-deployment

This PR makes able the deployment from Git Source avoiding the need of upload the files to Vercel.

It works passing a new parameter called `git_source`, it is a map and it looks like this:
```
resource "vercel_deployment" "deployment" {
  team_id = var.vercel_team_id
  project_id = var.vercel_project_id
  git_source = {
    ref = var.git_branch
    repo_id = var.fe_github_repo_id
    type = "github"
  }
}
```

Here is an example of output, it just works:
```
vercel_deployment.deployment: Creating...
vercel_deployment.deployment: Still creating... [10s elapsed]
vercel_deployment.deployment: Still creating... [20s elapsed]
vercel_deployment.deployment: Still creating... [30s elapsed]
vercel_deployment.deployment: Still creating... [40s elapsed]
vercel_deployment.deployment: Still creating... [50s elapsed]
vercel_deployment.deployment: Still creating... [1m0s elapsed]
vercel_deployment.deployment: Creation complete after 1m6s [id=dpl_EQBQgeWeo5PXesr1MLSjrQC9MRrZ]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```

### DIFF
--- a/client/deployment_create.go
+++ b/client/deployment_create.go
@@ -20,6 +20,11 @@ type DeploymentFile struct {
 	Sha  string `json:"sha"`
 	Size int    `json:"size"`
 }
+type GitSource struct {
+	RepoId string `json:"repoId"`
+	Ref    string `json:"ref"`
+	Type   string `json:"type"`
+}
 
 // CreateDeploymentRequest defines the request the Vercel API expects in order to create a deployment.
 type CreateDeploymentRequest struct {
@@ -35,6 +40,7 @@ type CreateDeploymentRequest struct {
 	Regions         []string               `json:"regions,omitempty"`
 	Routes          []interface{}          `json:"routes,omitempty"`
 	Target          string                 `json:"target,omitempty"`
+	GitSource       *GitSource             `json:"gitSource,omitempty"`
 }
 
 // DeploymentResponse defines the response the Vercel API returns when a deployment is created or updated.
@@ -59,15 +65,16 @@ type DeploymentResponse struct {
 	Build struct {
 		Environment []string `json:"env"`
 	} `json:"build"`
-	AliasAssigned    bool    `json:"aliasAssigned"`
-	ChecksConclusion string  `json:"checksConclusion"`
-	ErrorCode        string  `json:"errorCode"`
-	ErrorMessage     string  `json:"errorMessage"`
-	ID               string  `json:"id"`
-	ProjectID        string  `json:"projectId"`
-	ReadyState       string  `json:"readyState"`
-	Target           *string `json:"target"`
-	URL              string  `json:"url"`
+	AliasAssigned    bool      `json:"aliasAssigned"`
+	ChecksConclusion string    `json:"checksConclusion"`
+	ErrorCode        string    `json:"errorCode"`
+	ErrorMessage     string    `json:"errorMessage"`
+	ID               string    `json:"id"`
+	ProjectID        string    `json:"projectId"`
+	ReadyState       string    `json:"readyState"`
+	Target           *string   `json:"target"`
+	URL              string    `json:"url"`
+	GitSource        GitSource `json:"gitSource"`
 }
 
 // IsComplete is used to determine whether a deployment is still processing, or whether it is fully done.
--- a/docs/resources/deployment.md
+++ b/docs/resources/deployment.md
@@ -63,13 +63,14 @@ resource "vercel_deployment" "example" {
 
 ### Required
 
-- `files` (Map of String) A map of files to be uploaded for the deployment. This should be provided by a `vercel_project_directory` or `vercel_file` data source.
 - `project_id` (String) The project ID to add the deployment to.
 
 ### Optional
 
 - `delete_on_destroy` (Boolean) Set to true to hard delete the Vercel deployment when destroying the Terraform resource. If unspecified, deployments are retained indefinitely. Note that deleted deployments are not recoverable.
 - `environment` (Map of String) A map of environment variable names to values. These are specific to a Deployment, and can also be configured on the `vercel_project` resource.
+- `files` (Map of String) A map of files to be uploaded for the deployment. This should be provided by a `vercel_project_directory` or `vercel_file` data source. Required if `git_source` is not set
+- `git_source` (Attributes) A map with the Git repo information. Required if `files` is not set (see [below for nested schema](#nestedatt--git_source))
 - `path_prefix` (String) If specified then the `path_prefix` will be stripped from the start of file paths as they are uploaded to Vercel. If this is omitted, then any leading `../`s will be stripped.
 - `production` (Boolean) true if the deployment is a production deployment, meaning production aliases will be assigned.
 - `project_settings` (Attributes) Project settings that will be applied to the deployment. (see [below for nested schema](#nestedatt--project_settings))
@@ -80,6 +81,16 @@ resource "vercel_deployment" "example" {
 - `domains` (List of String) A list of all the domains (default domains, staging domains and production domains) that were assigned upon deployment creation.
 - `id` (String) The ID of this resource.
 - `url` (String) A unique URL that is automatically generated for a deployment.
+
+<a id="nestedatt--git_source"></a>
+### Nested Schema for `git_source`
+
+Optional:
+
+- `ref` (String) Branch or commit hash
+- `repo_id` (String) Frontend git repo ID
+- `type` (String) Type of git repo, supported values are: github
+
 
 <a id="nestedatt--project_settings"></a>
 ### Nested Schema for `project_settings`
--- a/vercel/resource_deployment.go
+++ b/vercel/resource_deployment.go
@@ -79,8 +79,8 @@ Once the build step has completed successfully, a new, immutable deployment will
 				Type:          types.BoolType,
 			},
 			"files": {
-				Description:   "A map of files to be uploaded for the deployment. This should be provided by a `vercel_project_directory` or `vercel_file` data source.",
-				Required:      true,
+				Description:   "A map of files to be uploaded for the deployment. This should be provided by a `vercel_project_directory` or `vercel_file` data source. Required if `git_source` is not set",
+				Optional:      true,
 				PlanModifiers: tfsdk.AttributePlanModifiers{tfsdk.RequiresReplace()},
 				Type: types.MapType{
 					ElemType: types.StringType,
@@ -88,6 +88,34 @@ Once the build step has completed successfully, a new, immutable deployment will
 				Validators: []tfsdk.AttributeValidator{
 					mapItemsMinCount(1),
 				},
+			},
+			"git_source": {
+				Description:   "A map with the Git repo information. Required if `files` is not set",
+				Optional:      true,
+				PlanModifiers: tfsdk.AttributePlanModifiers{tfsdk.RequiresReplace()},
+				Attributes: tfsdk.SingleNestedAttributes(map[string]tfsdk.Attribute{
+					"repo_id": {
+						Required:      true,
+						PlanModifiers: tfsdk.AttributePlanModifiers{tfsdk.RequiresReplace()},
+						Type:          types.StringType,
+						Description:   "Frontend git repo ID",
+					},
+					"ref": {
+						Required:      true,
+						PlanModifiers: tfsdk.AttributePlanModifiers{tfsdk.RequiresReplace()},
+						Type:          types.StringType,
+						Description:   "Branch or commit hash",
+					},
+					"type": {
+						Required:      true,
+						PlanModifiers: tfsdk.AttributePlanModifiers{tfsdk.RequiresReplace()},
+						Type:          types.StringType,
+						Description:   "Type of git repo, supported values are: github",
+						Validators: []tfsdk.AttributeValidator{
+							stringOneOf("github", "gitlab", "bitbucket", "custom"),
+						},
+					},
+				}),
 			},
 			"project_settings": {
 				Description:   "Project settings that will be applied to the deployment.",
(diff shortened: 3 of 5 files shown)


WCMinor:
After few tests this just worked well all the times, we're using our fork in prod, so please feel free to merge if you want/like or to propose changes. Thanks for your job guys!

WCMinor:
Could you please review and merge if you think it is ok?

docs/resources/deployment.md
```diff
@@ -63,12 +63,13 @@ resource "vercel_deployment" "example" {
...
 ### Required
 
-- **files** (Map of String) A map of files to be uploaded for the deployment. This should be provided by a `vercel_project_directory` or `vercel_file` data source.
 - **project_id** (String) The project ID to add the deployment to.
 
 ### Optional
 
 - **environment** (Map of String) A map of environment variable names to values. These are specific to a Deployment, and can also be configured on the `vercel_project` resource.
+- **files** (Map of String) A map of files to be uploaded for the deployment. This should be provided by a `vercel_project_directory` or `vercel_file` data source.
+- **git_source** (Attributes) A map with the Git repo information (see [below for nested schema](#nestedatt--git_source))
```
dglsparsons, line 72: I think we should probably validate that one of `files` or `git_source` is present.

docs/resources/deployment.md
```diff
@@ -80,6 +81,16 @@ resource "vercel_deployment" "example" {
...
 - **url** (String) A unique URL that is automatically generated for a deployment.
 
+<a id="nestedatt--git_source"></a>
+### Nested Schema for `git_source`
+
+Optional:
+
+- **ref** (String) Branch or commit hash
+- **repo_id** (String) Frontend git repo ID
+- **type** (String) Type of git repo, supported values are: github
```
dglsparsons, line 91: How are these all optional?

vercel/resource_deployment.go
```diff
@@ -89,6 +89,34 @@ Once the build step has completed successfully, a new, immutable deployment will
...
+						Required:      true,
+						PlanModifiers: tfsdk.AttributePlanModifiers{tfsdk.RequiresReplace()},
+						Type:          types.StringType,
+						Description:   "Branch or commit hash",
+					},
+					"type": {
+						Required:      true,
+						PlanModifiers: tfsdk.AttributePlanModifiers{tfsdk.RequiresReplace()},
+						Type:          types.StringType,
+						Description:   "Type of git repo, supported values are: github",
```
dglsparsons, line 113: Is this a restriction of the deplyoment API, or is github the only supported value?

vercel/resource_deployment_model.go
```diff
@@ -45,6 +53,16 @@ func setIfNotUnknown(m map[string]interface{}, v types.String, name string) {
...
 
+// toRequest takes the input of GitSource and converts them into the required
+// format for a CreateDeploymentRequest.
+func (g *GitSource) toRequest() client.GitSource {
+	var gs client.GitSource
+	gs.Ref = g.Ref.Value
+	gs.RepoId = g.RepoId.Value
+	gs.Type = g.Type.Value
+	return gs
+}
```
dglsparsons, line 64: Why not: 
```
return client.GitSource{
   Ref: g.Ref.Value,
   RepoId: g.RepoId.Value, 
   Type: g.Type.Value,
}
```

vercel/resource_deployment_model.go
```diff
@@ -20,6 +20,13 @@ type ProjectSettings struct {
 	RootDirectory   types.String `tfsdk:"root_directory"`
 }
 
+// GitSource represents the Git integration source for a deployment.
+type GitSource struct {
+	RepoId types.String `tfsdk:"repo_id"`
+	Ref    types.String `tfsdk:"ref"`
+	Type   types.String `tfsdk:"type"`
+}
+
```
dglsparsons, line 29: I think we'd also want to add tests / documentation for this.

vercel/resource_deployment.go
WCMinor: I thought `github` was the only supported value but it looks like they support more in the API specification.
updating

docs/resources/deployment.md
WCMinor: I set them as required, let me check why they appear as optional in the docs

docs/resources/deployment.md
WCMinor: on it

docs/resources/deployment.md
WCMinor: I was investigating a bit and could not find why this happens. tfplugindocs just renders them as optional even when they are set as required. if happens the same [here](https://github.com/vercel/terraform-provider-vercel/blob/main/docs/resources/deployment.md#read-only)

vercel/resource_deployment.go
WCMinor: changed

docs/resources/deployment.md
dglsparsons: Ahh interesting. I guess it's because the parent property is optional then.

vercel/resource_deployment_model.go
WCMinor: working on these

WCMinor:
I think the tests are not passing because the PR is coming from an external repo and that mess the GH secrets maybe? the tests work locally on my end

docs/resources/deployment.md
WCMinor: this has been done in https://github.com/vercel/terraform-provider-vercel/pull/21/files#diff-03202b30898c33935f6776c185f1dd43d67d1d5c6e81fd443eb3a88bb45f4154R212-R220

vercel/resource_deployment_model.go
```diff
@@ -187,5 +205,19 @@ func convertResponseToDeployment(response client.DeploymentResponse, plan Deploy
 		PathPrefix:      fillStringNull(plan.PathPrefix),
 		ProjectSettings: plan.ProjectSettings.fillNulls(),
 		DeleteOnDestroy: plan.DeleteOnDestroy,
+		GitSource:       plan.GitSource,
+	}
+}
+
+// checkMutualyExclusiveAttributes checks whether git_source and files are not set at the same time
+func (d *Deployment) checkMutualyExclusiveAttributes() error {
```
dglsparsons, line 213: I think we can do this as a validation instead, but I'll merge the changes and fix this up retrospectively.

WCMinor: that makes sense

WCMinor: just to clarify, are you merging this or you require more changes?

client/deployment_create.go
```diff
@@ -20,6 +20,11 @@ type DeploymentFile struct {
 	Sha  string `json:"sha"`
 	Size int    `json:"size"`
 }
+type GitSource struct {
+	RepoId string `json:"repoId"`
```
dglsparsons, line 24: Digging into this, where does the `repoId` come from? 
I can see in the example that you have it as a number. 🤔. I wonder if we need some data source to expose this so it's more usable?

dglsparsons: Okay, so `repoId` looks like it is exposed through a project. I think it could be exposed from both the `vercel_project` resource and the `vercel_project` data source. 
This would mean you can use it like: 
```terraform
data "vercel_project" "my_project" {
   name = "my-project"
}

resource "vercel_deployment" "example" {
  team_id = var.vercel_team_id
  project_id = data.vercel_project.my_project.id
  git_source = {
    ref = "main" // or any other branch you fancy. 
    repo_id = data.vercel_project.my_project.git_repository.repo_id
    type = data.vercel_project.my_project.git_repository.type
  }
}
```

I think this would be more preferable than having to dig out a repoId. What do you think @WCMinor?

client/deployment_create.go
```diff
@@ -59,15 +65,16 @@ type DeploymentResponse struct {
...
+	AliasAssigned    bool      `json:"aliasAssigned"`
+	ChecksConclusion string    `json:"checksConclusion"`
+	ErrorCode        string    `json:"errorCode"`
+	ErrorMessage     string    `json:"errorMessage"`
+	ID               string    `json:"id"`
+	ProjectID        string    `json:"projectId"`
+	ReadyState       string    `json:"readyState"`
+	Target           *string   `json:"target"`
+	URL              string    `json:"url"`
+	GitSource        GitSource `json:"git_source"`
```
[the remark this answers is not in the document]
WCMinor: mmm, as per their [recommendations](https://www.terraform.io/plugin/sdkv2/best-practices/naming) it looks fine:
> Attribute names within Terraform configuration blocks are conventionally named as all-lowercase with underscores separating words, as shown above.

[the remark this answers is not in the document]
WCMinor: feel free to change it if you think it is more consistent

client/deployment_create.go
dglsparsons: That's true that you can find it out that way, but I think it's somewhat surprising to see (and results in a magic number). I think if we had the data source, or exposed it via a project it would be more intuitive on how to use the two together to trigger a deployment.

WCMinor: true, should I create an issue for that?

client/deployment_create.go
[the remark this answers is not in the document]
WCMinor: you're right, I just tested it
```
    },
    "gitSource": {
        "repoId": 449314886,
        "ref": "dario-terraform-deployment-test",
        "type": "github",
        "sha": "f4f09e995618dea3cd77e39a2cb710628ae03228"
```
changed

client/deployment_create.go
dglsparsons: Okay, so the create-deployment API looks a tad more complicated as it supports a few variations of `type (github)/ref/repoId` or `type (github)/ref/org/repo` or `type (gitlab) /ref/projectId` or `type (bitbucket)/ref/repoUuid` or `type (bitbucket)/ref/owner/slug`.

dglsparsons: I'll give this some thought and try to get back to you earlier next week.

dglsparsons: I'm curious on the need part - how do you currently deploy (prior to terraform)?

WCMinor: we build and install the provider on the deployment ci pipeline and change the Terraform configuration to use the local plugin instead of the public provider. Not ideal.

WCMinor: prior to terraform we used vercel cli directly on github actions

WCMinor: then we decided to implement a deployment using the vercel API directly, which quickly turned into just fork and use your provider which already does 90% of what we need. BTW we will need in the future to make dns aliases for deployments as well, not sure if this is already implemented in the provider, happy to help there too

dglsparsons: I have some work in progress for DNS records, assuming this is what you mean? The battle on my part is just getting time to do everything I intend to 😁. It is possible to use a deployment with `files` for your use case? I'm not 100% clear on the need for repo/commit IDs (could be missing something here though).

WCMinor: we need to deploy from a git repo, so that vercel clones the repo and deploy it on their side. Using `files` I need to have the files locally, we are not deploying `example_app` from the `example_app` git repo, but rather we are deploying `example_app1`, `example_app2`, `example_app3` ... from a `devops_terraform` git repo

client/deployment_create.go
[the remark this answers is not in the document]
dglsparsons: Hooray. We should probably set this back onto the terraform state, rather than using the value from `plan` in that case (so it's readable). However... I think we probably need to fix the `repoId` issue with different git providers first.

client/deployment_create.go
dglsparsons: Ahh okay, I understand. It's a pretty unusual workflow but I guess it makes sense. :) I will try to prioritise getting this in place. It's just tricky supporting the different git providers.

dglsparsons: Out of curiosity have you considered putting them in a single repository, behind something like https://turborepo.org/ ?

WCMinor: we're actually moving on from a monorepo to a distributed system to allow each team to manage their own code as the teams get bigger

dglsparsons: Makes sense! As an update on this issue - I've been having a play around, and I'm coming to the belief that the only thing needed is a `ref` field on the `vercel_deployment` resource. 

For example 
```terraform
   name = "my-test-project" 
   git_repository = {
       type = "github" 
       repo = "acme/example"
   }
}

# Deploy the main branch of acme/example. 
resource "vercel_deployment" "example" {
    ref        = "main" # or a commit ID. 
}
```

The deployment resource can, itself, look up the relevant repoId / equivalent bitbucket/gitlab parameters, and pass these onto the deployment creation. 

You could even use it with a data-source to pull a project by a name: 
```terraform 
data "vercel_project" "example" {
    name = "my-test-project" 
}

resource "vercel_deployment" "example" {
    project_id = data.vercel_project.example.id
    ref        = "main" 
}
```

We already have a hard requirement that for a deployment to be created, a project must already exist. I think using the git-connection this way would work nicely, and also prevents any confusing fields being presented back out of the provider (like repoId or projectUuid or workspaceUuid). What do you think @WCMinor? Would this work for your use-case? 

With my playing around I think i'm fairly close to a working version, so if this works for you, I should be able to have a PR up soonish.

WCMinor: This looks great yeah, much easier to understand and to configure.

Funny enough, if you pass a public github repo (not configured in your project) to the deploy API , it works, it deploys a project based on such public repo. That is the only case in which it would make sense to use a standalone `git_source` parameter. Honestly, I think that functionality is kinda out of the scope of this provider as it is not really documented by Vercel.

dglsparsons: Yeah. It's probably more a result of accidental API design than anything else 😓

dglsparsons:
Closing this in favour of #29

PR #22, opened 2022-04-25
DeleteDeployment: Create a delete deployment method

WCMinor:
For us it is important to delete deployments for terraform consistency

The deletion of the resource worked fine in our tests:
```
Terraform will perform the following actions:

  # vercel_deployment.deployment will be destroyed
  - resource "vercel_deployment" "deployment" {
      - domains    = [
          - "avocado-git-dario-terraform-deployment-test.pollen.run",
        ] -> null
      - git_source = {
          - ref     = "dario-terraform-deployment-test" -> null
          - repo_id = "449314886" -> null
          - type    = "github" -> null
        }
      - id         = "dpl_9gBRCvWiw8ViYzAawNypNfNrLtK2" -> null
      - production = false -> null
      - project_id = "prj_xxxxxxxxxxxxxxxxt" -> null
      - team_id    = "team_xxxxxxxxxxxxxxxxt" -> null
      - url        = "avocado-ibsmgnaua.pollen.run" -> null
    }

Plan: 0 to add, 0 to change, 1 to destroy.

───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

Saved the plan to: /tmp/plan.out

To perform exactly these actions, run the following command to apply:
    terraform apply "/tmp/plan.out"
~/dev/pollen/tf_module_vercel_deployment main* ❯ terraform apply /tmp/plan.out                                        
╷
│ Warning: Provider development overrides are in effect
│ 
│ The following provider development overrides are set in the CLI configuration:
│  - vercel/vercel in /terraform-provider-vercel
│ 
│ The behavior may therefore not match any released version of the provider and applying changes may cause the state to become incompatible with published releases.
╵
vercel_deployment.deployment: Destroying... [id=dpl_9xxxxxxxxxxxxxxxxtK2]
vercel_deployment.deployment: Destruction complete after 2s

Apply complete! Resources: 0 added, 0 changed, 1 destroyed.
```
The deployment was gone from the vercel console.

If deleted manually in the middle of the operation it fails as expected:
```
 ❯ terraform apply /tmp/plan.out                                        
╷
│ Warning: Provider development overrides are in effect
│ 
│ The following provider development overrides are set in the CLI configuration:
│  - vercel/vercel in /terraform-provider-vercel
│ 
│ The behavior may therefore not match any released version of the provider and applying changes may cause the state to become incompatible with published releases.
╵
vercel_deployment.deployment: Destroying... [id=dpl_5xxxxxxxxxxxxxxx]
╷
│ Error: Error deleting deployment
│ 
│ Could not delete deployment xxxxx-iuyda4x0l.pollen.run, unexpected error: not_found - Not found
╵
~/tf_module_vercel_deployment main* ❯ terraform plan --var-file=params.tfvars --out /tmp/plan.out --destroy
╷
│ Warning: Provider development overrides are in effect
│ 
│ The following provider development overrides are set in the CLI configuration:
│  - vercel/vercel in /terraform-provider-vercel
│ 
│ The behavior may therefore not match any released version of the provider and applying changes may cause the state to become incompatible with published releases.
╵
vercel_deployment.deployment: Refreshing state... [id=dpl_xxxxxxxxxxxy]

No changes. No objects need to be destroyed.
```

### DIFF
--- a/client/deployment_delete.go
+++ b/client/deployment_delete.go
@@ -1,0 +1,45 @@
+package client
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+)
+
+// DeleteDeploymentResponse defines the response the Vercel API returns when a deployment is deleted.
+type DeleteDeploymentResponse struct {
+	State string `json:"state"`
+	UID   string `json:"uid"`
+}
+
+// DeleteDeployment deletes a deployment within Vercel.
+func (c *Client) DeleteDeployment(ctx context.Context, deploymentID string, teamID string) (r DeleteDeploymentResponse, err error) {
+	url := fmt.Sprintf("%s/v13/deployments/%s", c.baseURL, deploymentID)
+	req, err := http.NewRequest(
+		"DELETE",
+		url,
+		nil,
+	)
+	if err != nil {
+		return r, err
+	}
+
+	// Add query parameters
+	q := req.URL.Query()
+	if teamID != "" {
+		q.Add("teamId", teamID)
+	}
+	req.URL.RawQuery = q.Encode()
+
+	tflog.Trace(ctx, "deleting deployment", map[string]interface{}{
+		"url": url,
+	})
+	err = c.doRequest(req, &r)
+	if err != nil {
+		return r, err
+	}
+
+	return r, nil
+}
--- a/vercel/resource_deployment.go
+++ b/vercel/resource_deployment.go
@@ -129,6 +129,11 @@ Once the build step has completed successfully, a new, immutable deployment will
 					},
 				}),
 			},
+			"delete_on_destroy": {
+				Description: "Set to true to delete the Vercel deployment when destroying the Terraform resource. Defaults to False",
+				Optional:    true,
+				Type:        types.BoolType,
+			},
 		},
 	}, nil
 }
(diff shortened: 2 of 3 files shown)


vercel/resource_deployment.go
```diff
@@ -308,7 +308,26 @@ func (r resourceDeployment) Update(ctx context.Context, req tfsdk.UpdateResource
 // Delete does nothing other than clear the existing terraform state for a Deployment.
 // This is done intentionally, as typically, Vercel users do not continually delete old Deployments.
 func (r resourceDeployment) Delete(ctx context.Context, req tfsdk.DeleteResourceRequest, resp *tfsdk.DeleteResourceResponse) {
-	tflog.Trace(ctx, "deleted deployment")
+	//err := r.p.client.DeleteProject(ctx, state.ID.Value, state.TeamID.Value)
```
dglsparsons, line 311: I'm hesitant to merge this just because the comment here (L309) kind of conflicts with the implementation. I'm really unsure whether people will want deployments deleted. 

AFAIK it's a fairly rare operation to perform on Vercel, with the majority of people leaving deployments around for later reference. It may be better to include some optional flag on the terraform state to indicate that the deployments should be deleted though, allowing this to be an opt-in behaviour.

vercel/resource_deployment.go
```diff
@@ -308,7 +308,26 @@ func (r resourceDeployment) Update(ctx context.Context, req tfsdk.UpdateResource
...
+	var state Deployment
+	diags := req.State.Get(ctx, &state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	dResp, err := r.p.client.DeleteDeployment(ctx, state.URL.Value, state.TeamID.Value)
```
dglsparsons, line 318: What's the rationale behind using URL over deploymentID?

vercel/resource_deployment.go
```diff
@@ -308,7 +308,26 @@ func (r resourceDeployment) Update(ctx context.Context, req tfsdk.UpdateResource
...
+			"Error deleting deployment",
+			fmt.Sprintf(
+				"Could not delete deployment %s, unexpected error: %s",
+				state.URL.Value,
+				err,
+			),
+		)
+		return
+	}
+	tflog.Trace(ctx, fmt.Sprintf("deleted deployment %s", dResp.UID))
```
dglsparsons, line 330: Prefer structured logging where possible:- 
```suggestion
	tflog.Trace(ctx, "deleted deployment", map[string]interface{}{
	    "deployment_id": dResp.UID
	})
```

client/deployment_delete.go
```diff
@@ -0,0 +1,45 @@
...
+		url,
+		nil,
+	)
+	if err != nil {
+		return r, err
+	}
+
+	// Add query parameters
+	q := req.URL.Query()
+	q.Add("teamId", teamID)
```
dglsparsons, line 32: I think teamID should be conditional, so this should only be added if there is a teamID present? Or does `req.URL.Query()` handle this?

dglsparsons (changes requested):

vercel/resource_deployment.go
WCMinor: I thought the Id was not included in the create deployment response, I just double checked and it does, so I'll change it

client/deployment_delete.go
WCMinor: you're right, changing this

vercel/resource_deployment.go
WCMinor: What about making it optional? based on an input parameter for example? We use vercel deployments to make very ephemeral e2e tests (along with other purposes) so we create a lot of ephemeral deployments. As a Devops engineer I'm kinda inclined to delete them.
I'll make this optional.

WCMinor: And the default will be to leave them deployed.

dglsparsons: That would be 👍

vercel/resource_deployment.go
WCMinor: changed

vercel/resource_deployment.go
WCMinor: I changed the behaviour to be optional and default to false

dglsparsons (approved):

PR #23, opened 2022-04-25
Adding Content-Type headers to POST and PATCH requests

ctgowrie:
While working on some API improvements, I noticed that requests from the terraform provider do not set the content-type header, so this adds Content-Type to all POST and PATCH requests where relevant.

- sets default content-type header as application/json for any doRequest calls with a body
- change to nil body for GET and DELETE requests (so we do not set content-type header)
- application/octet-stream for file upload requests

### DIFF
--- a/client/deployment_create.go
+++ b/client/deployment_create.go
@@ -154,6 +154,8 @@ func (c *Client) CreateDeployment(ctx context.Context, request CreateDeploymentR
 	if err != nil {
 		return r, err
 	}
+	
+	req.Header.Set("Content-Type", "application/json")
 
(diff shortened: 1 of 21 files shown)


dglsparsons:
Hey @ctgowrie. Thanks for this - look like a nice simple improvement (and well spotted). 
I wonder if we need a helper function so this is more reliable / less copy-paste code. I'll merge this in the meantime though, as it's a nice improvement.

ctgowrie:
@dglsparsons Thanks! I can move the `application/json` header setting into doRequest `if req.Body != nil && req.Header.Get("Content-Type") == ""` pretty easily if you'd prefer

dglsparsons (approved):

PR #27, opened 2022-05-04
Allow updating Deployment delete_on_destroy field

dglsparsons:


vercel/resource_deployment.go
```diff
@@ -303,15 +303,40 @@ func (r resourceDeployment) Read(ctx context.Context, req tfsdk.ReadResourceRequ
...
+// of setting terraform state.
 func (r resourceDeployment) Update(ctx context.Context, req tfsdk.UpdateResourceRequest, resp *tfsdk.UpdateResourceResponse) {
-	// Nothing to do here - we can't update deployments
+	var plan Deployment
+	diags := req.Plan.Get(ctx, &plan)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		resp.Diagnostics.AddError(
+			"Error getting deployment plan",
+			"Error getting deployment plan",
```
craigandrews, line 316: nothing else worth adding here?

vercel/resource_deployment.go
```diff
@@ -303,15 +303,40 @@ func (r resourceDeployment) Read(ctx context.Context, req tfsdk.ReadResourceRequ
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Copy over the planned field only
+	state.DeleteOnDestroy = plan.DeleteOnDestroy
+	diags = resp.State.Set(ctx, state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
```
craigandrews, line 333: two bare returns that don't surface the error

vercel/resource_deployment.go
dglsparsons: I copy-pasted from the `Create` function. Might have a play and see what the output looks like 🤔. I suspect it might be one of those errors that's really hard to actually see, just because the schema validation kicks in before this point.

craigandrews: yeah it's worth figuring out if the errors raised here are actually useful for output, or if they're pretty much just flags to say "stop now, it didn't work"

craigandrews (approved):
Approved on the proviso that error outputs are investigated

PR #29, opened 2022-05-10
Add ability to create Deployment from git ref

dglsparsons:
- Refactor test to use env variables for repositories


dglsparsons:
FYI @WCMinor :)

WCMinor:
I'm testing your code, if I pass the value of `ref` as a hardcoded config, like in
```
ref     = "main"
```
the validation works fine. However if I pass it as a variable, as in
```
ref     = var.git_branch
```
It does not work, is throws:
```
A Deployment must have either `ref` or `files` specified
```

Of course the variable `git_branch` has value in my template. This is a behaviour that I found days ago, that's why I originally implemented it in the apply state (hacky).

dglsparsons:
@WCMinor That's a good catch! I think it's because I'm checking `Value` rather than `Null`. I'll update that! :)

client/deployment_create.go
```diff
@@ -136,10 +149,52 @@ func (e MissingFilesError) Error() string {
...
+		return gitSource{
+			Type: "github",
+			Org:  project.Link.Org,
+			Repo: project.Link.Repo,
+			Ref:  ref,
+		}, nil
+	case "gitlab":
+		return gitSource{
+			ProjectID: project.Link.ProjectID,
+			Type:      "gitlab",
```
ecklf, line 172: Maybe sort these across all the different git sources so it looks cleaner?

README.md
```diff
@@ -37,9 +37,12 @@ In order to test the provider, you can simply run `task test`.
 
 _Note:_ This runs acceptance tests that will create real resources. You should expect that the full acceptance test suite will take some time to run.
 
-The acceptance tests require a `VERCEL_API_TOKEN` (which can be generated [here](https://vercel.com/account/tokens), and
-a `VERCEL_TERRAFORM_TESTING_TEAM` (which should be a Vercel team_id where resources can be created and destroyed)
-environment variable set.
+The acceptance tests require a few environment variables to be set:
+* `VERCEL_API_TOKEN` - this can be generated [here](https://vercel.com/account/tokens)
+* `VERCEL_TERRAFORM_TESTING_TEAM` - a Vercel team_id where resources can be created and destroyed
+* `VERCEL_TERRAFORM_TESTING_GITHUB_REPO` - a github repository in the form 'org/repo' that can be used to trigger deployments
```
ecklf, line 43: Since we capitalize the Vercel brand name we should also do it for others?
`GitHub`, `GitLab`, `Bitbucket`, etc.

PR #58, opened 2022-08-23
Support individual environment resources

ecklf:


client/environment_variables_get.go
```diff
@@ -32,3 +32,25 @@ func (c *Client) getEnvironmentVariables(ctx context.Context, projectID, teamID
 	err = c.doRequest(req, &envResponse)
 	return envResponse.Env, err
 }
+
+func (c *Client) GetEnvironmentVariable(ctx context.Context, projectID, teamID, envID string) (e EnvironmentVariable, err error) {
+	url := fmt.Sprintf("%s/v1/projects/%s/env/%s?decrypt=true", c.baseURL, projectID, envID)
```
dglsparsons, line 37: can we check if this needs the `decrypt=true` parameter?

client/project_create.go
```diff
@@ -27,6 +27,17 @@ type EnvironmentVariable struct {
 	ID        string   `json:"id,omitempty"`
 }
 
+// type?: "system" | "secret" | "encrypted" | "plain"
```
dglsparsons, line 30: remove this :)

vercel/resource_project_environment_variable.go
```diff
@@ -0,0 +1,239 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/provider"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/vercel/terraform-provider-vercel/client"
+)
+
+type resourceProjectEnvironmentVariableType struct{}
```
dglsparsons, line 16: you need to wire this in to `provider.go`

vercel/resource_project_environment_variable.go
```diff
@@ -0,0 +1,239 @@
...
+				Type:          types.StringType,
+			},
+			"value": {
+				Required:    true,
+				Description: "The value of the environment variable.",
+				Type:        types.StringType,
+			},
+			"git_branch": {
+				Optional:    true,
+				Description: "The git branch of the environment variable",
```
dglsparsons, line 46: We are inconsistent on whether the Description ends with a full-stop or not. Sorry if you've copy pasted that from existing code, but lets make it consistent :)

vercel/resource_project_environment_variable.go
```diff
@@ -0,0 +1,239 @@
...
+
+type resourceProjectEnvironmentVariableType struct{}
+
+// GetSchema returns the schema information for an project environment variable resource.
+func (r resourceProjectEnvironmentVariableType) GetSchema(_ context.Context) (tfsdk.Schema, diag.Diagnostics) {
+	return tfsdk.Schema{
+		Description: `
+Provides a Project Environment Variable resource.
+
+An Project Environment Variable resource allows to add an environment variable to a Vercel project `,
```
dglsparsons, line 24: `allows to add` doesn't make sense gramatically here. Could we also link to the Vercel documentation for Environment Variables? https://vercel.com/docs/concepts/projects/environment-variables

vercel/resource_project_environment_variable.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	out, err := r.p.client.GetEnvironmentVariable(ctx, state.ProjectID.Value, state.TeamID.Value, state.ID.Value)
+
+	if client.NotFound(err) {
+		resp.State.RemoveResource(ctx)
+		return
+	}
+
+	if err != nil {
```
dglsparsons, line 141: nit: 

I'd probably remove the whitespace between the operation and it's error handling. This just gives a subtle hint that they are related as it keeps them logically grouped.
So: 
```suggestion
	out, err := r.p.client.GetEnvironmentVariable(ctx, state.ProjectID.Value, state.TeamID.Value, state.ID.Value)
	if client.NotFound(err) {
		resp.State.RemoveResource(ctx)
		return
	}
	if err != nil {
```

vercel/resource_project_environment_variable.go
```diff
@@ -0,0 +1,239 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	var state ProjectEnvironmentVariable
+	diags = req.State.Get(ctx, &state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
```
dglsparsons, line 182: You don't use `state` after reading it (this is only necessary if we want to work out what has changed). Since we just upsert I think you only care about the `plan` here.

vercel/resource_project_environment_variable_model.go
```diff
@@ -0,0 +1,53 @@
+package vercel
+
+import (
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/vercel/terraform-provider-vercel/client"
+)
+
+// Alias represents the terraform state for an alias resource.
+// EnvironmentItem reflects the state terraform stores internally for a project's environment variable.
+type ProjectEnvironmentVariable struct {
```
dglsparsons, line 10: Comments mismatch here.

client/environment_variables_get.go
ecklf: Yeah we don't need that

dglsparsons (approved):

PR #59, opened 2022-08-23
Prevent project field coercion on creation

### LINKED ISSUE #56: Error: Provider produced inconsistent result after apply
Hi,

When trying to perform a `terraform apply` with a new Vercel Project with `public_source = false`, I got the following error:
Screen Shot 2022-08-22 at 14 03 00

Below is the resource that was attempted to be created (environment has a list of vars not included):
```
resource "vercel_project" "ecom-storefront" {
  name      = "ecom-storefront-1"
  framework = "nextjs"
  git_repository = {
    type = "github"
    repo = "a-github-repo"
  }
  team_id                    = "a-team-id"
  serverless_function_region = "iad1"
  install_command            = "npm install -g npm-cli-login && npm-cli-login && npm install"
  public_source              = false
  environment = []
}
```

Steps Taken to Debug:

1. Verify `terraform fmt -check`
2. Verify `terraform validate`
3. Verify latest Provider version (using `~> 0.7`)
4. Retry `terraform apply` (destroy works as expected, creation fails again)

Details from `terraform version`:
```
Terraform v1.2.6
on darwin_amd64
+ provider registry.terraform.io/hashicorp/aws v4.27.0
+ provider registry.terraform.io/vercel/vercel v0.7.0
```

dglsparsons:
In the Vercel API the following fields are coerced to null during project creation
if they are falsy in the create-project request:
* BuildCommand
* DevCommand
* InstallCommand
* OutputDirectory
* PublicSource

This causes an issue when they are specified, but falsy, as the
terraform configuration explicitly sets a value for them, but the Vercel
API returns a different value. This causes an inconsistent plan error.

We avoid this issue by choosing to use values from the terraform state,
but only if they are _explicitly stated_ *and* they are _falsy_ values
*and* the response value was Null. This is important as drift detection
would fail to work if the value was always selected.

Closes #56


dglsparsons:
2 extra changes I'll make before this gets merged: 
- add a test 
- improve the documentation of public_source.

ecklf (approved):

PR #69, opened 2022-09-01
Automatically gather serverless_function_regions

dglsparsons:
Change the validation on the serverless_function_region field of a
project to automatically read them, rather than using a hard-coded list.
This will ease the maintenance of the provider as new regions will
automatically become available.

I've also taken some time to improve the error messages returned when an
invalid serverless_function_region or framework is specified on a
project. I've added tests for these too.


kakadiadarpan (approved):
Nice! 🚀

Issue #71, opened 2022-09-08
project_environment_variable values should be sensitive


viki53:
Hi,

I've started to use your provider earlier today and I'm having a few issues and questions.

My main issue is the difference of behavior between the `vercel_project.environment` block and the `vercel_project_environment_variable` resource: the first one marks all values as sensitive (which is good for security) but the later doesn't so I risk having values being output when I don't want them displayed.

I also find weird having to specify an id is the `vercel_project.environment` blocks, but I won't be using them as I much prefer having dedicated resources anyway.

Also, now that this new `vercel_project_environment_variable` resource exists, not specifying the `environment` property in a `vercel_project` shouldn't cause an update of the resource if I never had it specified before: this makes me think that every time I'll update a project property I risk breaking all my env vars and having to re-import them via the `vercel_project_environment_variable` resources, which I define in different TF runs (kind of like the architecture in #47).

---

Update: I tried applying the changes on my project and it did break all the env vars (including those that were defined in the same run using `vercel_project_environment_variable` resources) as I had feared… I had to re-add all of them.   
**This is not OK for a stable deployment process!**

dglsparsons:
Hey @viki53, thanks for raising this. I'll try to address your points one at a time here: 

> My main issue is the difference of behavior between the vercel_project.environment block and the vercel_project_environment_variable resource: the first one marks all values as sensitive (which is good for security) but the later doesn't so I risk having values being output when I don't want them displayed.

That's a good catch - the `project_environment_variable` resource is a recent addition. It looks like the `sensitive` field has been accidentally omitted. I'll add this and ship a new provider version. 

> I also find weird having to specify an id is the vercel_project.environment blocks, but I won't be using them as I much prefer having dedicated resources anyway.

The `id` field is a computed value so not something you need to specify. It's used internally for updates, but hopefully isn't something you need to think about ever. 

> not specifying the environment property in a vercel_project shouldn't cause an update of the resource if I never had it specified before this makes me think that every time I'll update a project property I risk breaking all my env vars and having to re-import them via the vercel_project_environment_variable resources

I'm a bit confused about what you mean here. If you've never specified the `environment` property, why would anything change? Are you able to share a reproduction of the issue you encountered here? 

> Update: I tried applying the changes on my project and it did break all the env vars (including those that were defined in the same run using vercel_project_environment_variable resources) as I had feared… I had to re-add all of them.

[...]

PR #72, opened 2022-09-09
Mark project_environment_variable value as sensitive

dglsparsons:
This prevents it being displayed in plan/apply output. 

Ref #71


kakadiadarpan (commented):
We also don't accept `type`(`system` | `secret` | `encrypted` | `plain`) param, let's add that as well?

kakadiadarpan (approved):

PR #77, opened 2022-09-15
Parse Gitlab repo slugs from the response URL rather than using Project

### LINKED ISSUE #76: "git_repository" can't setup project when project name not lowercase
I have problem when setup vercel using terraform.

Condition:
my repository path in gitlab **`user-name/repository-project`** and project name in gitlab **`Repository Project`**, when i'm run **`terraform plan`** its doesn't have any problem, but when run **`terraform apply`** a problem message like this:
```
Plan: 1 to add, 0 to change, 1 to destroy.
vercel_project.example: Destroying... [id=prj_KAhhyBtevEDJUxxxxxxx]
vercel_project.example: Destruction complete after 1s
vercel_project.example: Creating...
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to vercel_project.example, provider "provider[\"registry.terraform.io/vercel/vercel\"]" produced an unexpected new value: .git_repository.repo: was
│ cty.StringVal("user-name/repository-name"), but now cty.StringVal("user-name/Repository Name").
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

I think the provider doesn't support when the format path in the repository is not same as project name in the repository or format only supports lowercase without whitespace, but when I change the project name only lowercase without whitespace and same with path in the repository it works.

This is terraform configuration file:
```
  name = "project"
  framework = "nextjs"

  git_repository = {
    repo = "user-name/repository-name"
    type = "gitlab"
  }
} 
```

dglsparsons:
Name

Closes #76

### DIFF
--- a/client/project_get.go
+++ b/client/project_get.go
@@ -15,8 +15,8 @@ type Repository struct {
 	Repo string
 }
 
-// getRepoNameFromURL is a helper method to extract the repo name from a gitlab URL.
-// This is necessary as Gitlab doesn't return the repository slug in the API response,
+// getRepoNameFromURL is a helper method to extract the repo name from a GitLab URL.
+// This is necessary as GitLab doesn't return the repository slug in the API response,
 // Because this information isn't present, the only way to obtain it is to parse the URL.
 func getRepoNameFromURL(url string) string {
 	url = strings.TrimSuffix(url, ".git")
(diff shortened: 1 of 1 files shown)


ecklf:
Looks good, but let's fix the inconsistent casing 👍🏻

ecklf (approved):

PR #78, opened 2022-09-19
Upgrade to terraform framework v13

dglsparsons:
- [V12 release notes](https://github.com/hashicorp/terraform-plugin-framework/releases/tag/v0.12.0)
- [V13 release notes](https://github.com/hashicorp/terraform-plugin-framework/releases/tag/v0.13.0)

### DIFF
--- a/vercel/data_source_alias.go
+++ b/vercel/data_source_alias.go
@@ -6,16 +6,44 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-framework/datasource"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
-	"github.com/hashicorp/terraform-plugin-framework/provider"
 	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/vercel/terraform-provider-vercel/client"
 )
 
-type dataSourceAliasType struct{}
+func newAliasDataSource() datasource.DataSource {
+	return &aliasDataSource{}
+}
+
+type aliasDataSource struct {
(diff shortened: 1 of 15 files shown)


vercel/resource_project.go
```diff
@@ -105,10 +132,9 @@ At this time you cannot use a Vercel Project resource with in-line ` + "`environ
 						Sensitive:   true,
 					},
 					"id": {
-						Description:   "The ID of the Environment Variable.",
-						Type:          types.StringType,
-						PlanModifiers: tfsdk.AttributePlanModifiers{resource.UseStateForUnknown()},
```
adriancooney, line 110: We don't need these modifiers anymore? :o

adriancooney (approved):
A lot cleaner! Great work.

vercel/data_source_alias.go
```diff
@@ -6,16 +6,44 @@ import (
...
+type aliasDataSource struct {
+	client *client.Client
+}
+
+func (d *aliasDataSource) Metadata(ctx context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_alias"
+}
+
+func (d *aliasDataSource) Configure(ctx context.Context, req datasource.ConfigureRequest, resp *datasource.ConfigureResponse) {
+	// Prevent panic if the provider has not been configured.
```
kakadiadarpan, line 28: 😄

kakadiadarpan (approved):

vercel/resource_project.go
dglsparsons: It's actually incorrect here - the project resource doesn't bother doing an update on the environment variables, and instead just deletes any that do not match the state after, and adds new ones. 

So we can't rely on stable IDs between updates for env variables.

PR #96, opened 2023-01-06
Omit `null` environment variables in vercel_deployment

### LINKED ISSUE #95: Optional environment variable using null value
I'd like to optionally declare an environment variable within `vercel_deployment` using a ternary/conditional expression by either setting a string or null.
For example:

```
environment = {
    MY_VAR = terraform.workspace == "default" ? "value" : null
}
```

This however throws an `unhandled null value` exception.

The only workaround right now is to instead set an empty string:

```
environment = {
    MY_VAR = terraform.workspace == "default" ? "value" : ""
}
```

I believe it would be a good addition also allow null values, and just ignore the `environment` entry then.

dglsparsons:
Creating a vercel_deployment and passing a `null` environment variable results in a conversion error.
```
An unexpected error was encountered trying to build a value. This is always
an error in the provider. Please report the following to the provider
developer:

Received null value, however the target type cannot handle null values. Use
the corresponding `types` package type, a pointer type or a custom type that
handles null values.

Path: ["BAZ"]
Target Type: string
Suggested `types` Type: basetypes.StringValue
Suggested Pointer Type: *string
```

We need to handle the case where the field value is `null`, which is currently omitted.

Since the API only accepts string values for enviroment variables, the best solution is to simply omit sending `null` to the API altogether.

Closes #95

### DIFF
--- a/vercel/resource_deployment_test.go
+++ b/vercel/resource_deployment_test.go
@@ -109,7 +109,8 @@ func TestAcc_DeploymentWithEnvironment(t *testing.T) {
 			{
 				Config: testAccDeploymentConfig(projectSuffix, teamIDConfig(), `environment = {
                     FOO = "baz",
-                    BAR = "qux"
+                    BAR = "qux",
+                    BAZ = null
                 }`),
 				Check: resource.ComposeAggregateTestCheckFunc(
 					testAccDeploymentExists("vercel_deployment.test", ""),
@@ -286,7 +287,7 @@ resource "vercel_deployment" "test" {
   project_id = vercel_project.test.id
   %[2]s
 
-  files       = data.vercel_prebuilt_project.test.output 
+  files       = data.vercel_prebuilt_project.test.output
   path_prefix = data.vercel_prebuilt_project.test.path
 }
 `, projectSuffix, teamID)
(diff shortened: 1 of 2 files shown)


vercel/resource_deployment_test.go
```diff
@@ -109,7 +109,8 @@ func TestAcc_DeploymentWithEnvironment(t *testing.T) {
 			{
 				Config: testAccDeploymentConfig(projectSuffix, teamIDConfig(), `environment = {
                     FOO = "baz",
-                    BAR = "qux"
+                    BAR = "qux",
+                    BAZ = null
```
adriancooney, line 113: Do we need to check that `BAZ` is not present on the output? I suppose this test would fail otherwise.

adriancooney (approved):

vercel/resource_deployment_test.go
dglsparsons: The test failed before the changes b/c of the deserialization issue. It's a bit rubbish but the test helpers don't give a nice way to check for a `null` value on the output. I think just checking that it doesn't cause the provider to panic is as good as we can do (which is the purpose of this).

ecklf (approved):

PR #98, opened 2023-01-11
fix the link to the docs

raphaelbadia:
I don't know much about terraform and I got confused when I clicked the link, took me a while to see the "documentation" tab on the right. Maybe this link should point directly to the docs ?


dglsparsons (approved):

dglsparsons:
Thanks for this - great improvement!

PR #119, opened 2023-06-14
Update team ID description in docs

kitfoster:
Currently it's unclear that the `team_id` needs to be specified for team resources which leads to confusing permission errors.


balazs4:
@dglsparsons I wonder if [task docs](https://github.com/vercel/terraform-provider-vercel/blob/69716e8e43025bbe254337129f081f4be0aacfbe/README.md?plain=1#L62-L65) is still the way because if I run it on `main`, there are a few parameters they became `Optional` instead of `Required`.

I would add the new `team_id` related comments to Description fields within the provider, and let `task docs` generate the docs for us. Any thoughts?

dglsparsons:
> @dglsparsons I wonder if [task docs](https://github.com/vercel/terraform-provider-vercel/blob/69716e8e43025bbe254337129f081f4be0aacfbe/README.md?plain=1#L62-L65) is still the way to generate docs because if I run it on main, there are a few parameters they became Optional instead of Required.

Interesting. It should be correct. I don't see any differences from rerunning `task docs` 🤔

dglsparsons (approved):

PR #145, opened 2023-11-21
Update Deployment Protection support

### LINKED ISSUE #144: vercel_authentication can't be disabled or configured correctly
Since https://github.com/vercel/terraform-provider-vercel/pull/143, existing projects are now having `vercel_authentication` added, to line up with the new Vercel defaults for projects. This is sensible, but the attributes available in the provider don't seem to allow for disabling the setting, which we need to do to maintain our existing project's configuration.

No matter how I update things, I am only able to achieve the below outcomes, implying that authentication will always be enabled:

```diff
+ vercel_authentication = {
  + protect_production = (true | false)
}
```

If the field must now be computed, it would be prudent to introduce a way to disable authentication altogether, like the console allows you to do.

The attributes also don't seem to match the potential configuration options available in Vercel, which are more akin to an enum selection, instead of boolean flags (as can be seen on the [changelog screenshot](https://vercel.com/changelog/deployment-protection-is-now-enabled-by-default-for-new-projects)). I wonder if an API like below would more closely align with Vercel:

```hcl
vercel_authentication = {
  enabled         = bool
  protection_type = "standard_protection" | "all_deployments" | "only_preview_deployments"
}
```

kitfoster:
- Add support for [Trusted IPs](https://vercel.com/docs/security/deployment-protection/methods-to-protect-deployments/trusted-ips)
- **[Breaking change]** Replace `protect_production` (boolean) with `deployment_type` (string) for [Password Protection](https://vercel.com/docs/security/deployment-protection/methods-to-protect-deployments/password-protection), [Vercel Authentication](https://vercel.com/docs/security/deployment-protection/methods-to-protect-deployments/vercel-authentication), and [Trusted IPs](https://vercel.com/docs/security/deployment-protection/methods-to-protect-deployments/trusted-ips)
- Make Vercel Authentication disable-able
- Update related tests & docs

Closes #144

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,6 +39,23 @@ jobs:
         run: task build
       - name: Format
         run: task lint
+
+  docs:
+    name: Docs Test
+    timeout-minutes: 5
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: "1.20"
+        id: go
+      - name: Install Task
+        uses: arduino/setup-task@v1
+        with:
+          version: "3.x"
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
       - name: Check if docs are up-to-date
         run: |
           task docs
@@ -48,7 +65,7 @@ jobs:
 
   test:
     name: Matrix Test
-    needs: [ build ]
+    needs: [build]
     timeout-minutes: 15
     strategy:
       max-parallel: 1
@@ -80,7 +97,7 @@ jobs:
         env:
           TF_ACC: "true"
           VERCEL_API_TOKEN: ${{ secrets.VERCEL_API_TOKEN }}
-          VERCEL_TERRAFORM_TESTING_TEAM: "team_Q8MH2GVRnqKLtxicP2HWPgLi"
+          VERCEL_TERRAFORM_TESTING_TEAM: "team_RvxIb1z0pi9RSsQ13p3ES4cK"
           VERCEL_TERRAFORM_TESTING_GITHUB_REPO: "dglsparsons/test"
           VERCEL_TERRAFORM_TESTING_GITLAB_REPO: "dglsparsons/test"
           VERCEL_TERRAFORM_TESTING_BITBUCKET_REPO: "dglsparsons-test/test"
--- a/README.md
+++ b/README.md
@@ -42,6 +42,8 @@ The acceptance tests require a few environment variables to be set:
 * `VERCEL_TERRAFORM_TESTING_TEAM` - a Vercel team_id where resources can be created and destroyed
 * `VERCEL_TERRAFORM_TESTING_GITHUB_REPO` - a GitHub repository in the form 'org/repo' that can be used to trigger deployments
 * `VERCEL_TERRAFORM_TESTING_BITBUCKET_REPO` - a Bitbucket repository in the form 'project/repo' that can be used to trigger deployments
+* `VERCEL_TERRAFORM_TESTING_GITLAB_REPO` - a GitLab repository in the form 'project/repo' that can be used to trigger deployments
+* `VERCEL_TERRAFORM_TESTING_DOMAIN` - a Vercel testing domain that can be used for testing
 
 ```sh
 $ task test
@@ -58,6 +60,22 @@ To run a specific set of tests, use the `-run` flag and specify a regex pattern 
 ```sh
 $ task test -- -run 'TestAcc_Project*'
 ```
+
+## Running the provider locally
+
+Set up a local override on your machine
+
+```sh
+$ task install
+```
+
+Build the provider
+
+```sh
+$ task build
+```
+
+Create a `main.tf` file on your machine and use the [terraform cli](https://developer.hashicorp.com/terraform/tutorials/aws-get-started/install-cli#install-terraform) to test
 
 ## Building The Documentation
 
--- a/README.md
+++ b/README.md
@@ -23,11 +23,29 @@ If you wish to work on the provider, you'll first need [Go](http://www.golang.or
 
 To compile the provider, run `task build`. This will build the provider and put the provider binary in the repository root.
 
-In addition, you can run `task install` to set up a developer overrides in your ~/.terraformrc. This will then allow you
-to use your locally built provider binary.
+```sh
+$ task build
+```
+
+In addition, you can run `task install` to set up a developer overrides in your ~/.terraformrc. This will then allow you to use your locally built provider binary.
+
+```sh
+$ task install
+```
+
+Create a `main.tf` file on your machine and use the [terraform cli](https://developer.hashicorp.com/terraform/tutorials/aws-get-started/install-cli#install-terraform) to test
+
+```sh
+$ terraform plan
+$ terraform apply
+```
 
 When you are finished using a local version of the provider, running `task uninstall` will remove _all_ developer
 overrides.
+
+```sh
+$ task uninstall
+```
 
 - HashiCorp - [Development Overrides for Provider developers](https://www.terraform.io/docs/cli/config/config-file.html#development-overrides-for-provider-developers).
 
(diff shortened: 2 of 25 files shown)


vercel/resource_project.go
```diff
@@ -176,21 +175,14 @@ At this time you cannot use a Vercel Project resource with in-line ` + "`environ
...
-						Optional:    true,
-						Computed:    true,
-						Default:     booldefault.StaticBool(false),
+					"deployment_type": schema.StringAttribute{
+						Required:      true,
+						Description:   "The deployment environment to protect. Must be one one `standard_protection`, `all_deployments`, or `only_preview_deployments`.",
+						PlanModifiers: []planmodifier.String{stringplanmodifier.UseStateForUnknown()},
+						Validators: []validator.String{
+							stringOneOf("standard_protection", "all_deployments", "only_preview_deployments"),
+						},
```
dglsparsons, line 185: I don't think this needs to be netsted under `vercel_authentication` any more? We could just have 
```hcl
    name = "some-project"
    vercel_authentication = "all_deployments"
}
```
WDYT?

dglsparsons: we can then just have a `Default` on it? :)

vercel/resource_project.go
```diff
@@ -176,21 +175,14 @@ At this time you cannot use a Vercel Project resource with in-line ` + "`environ
 			"vercel_authentication": schema.SingleNestedAttribute{
 				Description: "Ensures visitors to your Preview Deployments are logged into Vercel and have a minimum of Viewer access on your team.",
 				Optional:    true,
-				Computed:    true,
```
dglsparsons, line 179: Computed was added just because this has a default value when a new project is created. Need to make sure you can omit this field and it works.

vercel/resource_project.go
kitfoster: I think we should make the terraform reflect the api structure closeishly, otherwise if vercel auth gets a new settings we'll need another breaking change

.github/workflows/test.yml
```diff
@@ -39,6 +39,23 @@ jobs:
...
+        uses: actions/setup-go@v2
+        with:
+          go-version: "1.20"
+        id: go
+      - name: Install Task
+        uses: arduino/setup-task@v1
+        with:
+          version: "3.x"
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
```
kitfoster, line 58: separated the docs test because it's annoying not seeing the tests result because the docs check failed. Also, I updated the docs and can't figure out why I'm getting an error

dglsparsons: I'll make it req. once it's merged.

client/deployment_protection.go
```diff
@@ -0,0 +1,28 @@
...
+	DeploymentType string `json:"deploymentType"`
+}
+
+type PasswordProtection struct {
+	DeploymentType string `json:"deploymentType"`
+}
+type PasswordProtectionWithPassword struct {
+	DeploymentType string `json:"deploymentType"`
+	Password       string `json:"password"`
+}
```
dglsparsons, line 13:
```suggestion
type PasswordProtection struct {
	DeploymentType string `json:"deploymentType"`
	Password       *string `json:"password,omitempty"`
}
```

dglsparsons: Maybe, but this seems fine.

kitfoster: The matching struct `PasswordProtectionWithPassword` with `tfsdk` can't use the null pointer, so I think it's easier to keep them aligned

vercel/resource_project_model.go
```diff
@@ -185,19 +185,32 @@ func (p *PasswordProtection) toUpdateProjectRequest() *client.PasswordProtection
 	}
 }
 
+type TrustedIpsAddress struct {
```
dglsparsons:
```suggestion
type TrustedIpAddress struct {
```

vercel/resource_project.go
```diff
@@ -206,11 +210,48 @@ At this time you cannot use a Vercel Project resource with in-line ` + "`environ
...
+						},
+					},
+					"protection_mode": schema.StringAttribute{
+						Optional:      true,
+						Computed:      true,
+						Default:       stringdefault.StaticString("additional"),
+						Description:   "Whether or not Trusted IPs is optional to access a deployment. Must be either `additional` or `exclusive`. `exclusive` is only available with Standalone Trusted IPs.",
+						PlanModifiers: []planmodifier.String{stringplanmodifier.UseStateForUnknown()},
+						Validators: []validator.String{
+							stringOneOf("additional", "exclusive"),
```
dglsparsons, line 253: We... probably need to document what `additional` and `exclusive` means here, since it differs in terminology to the UI quite a lot.

vercel/resource_project.go
```diff
@@ -206,11 +210,48 @@ At this time you cannot use a Vercel Project resource with in-line ` + "`environ
...
+						Description:   "The allowed IP addressses and CIDR ranges with optional descriptions.",
+						Required:      true,
+						PlanModifiers: []planmodifier.Set{setplanmodifier.UseStateForUnknown()},
+						ElementType: types.ObjectType{
+							AttrTypes: map[string]attr.Type{
+								"value": types.StringType,
+								"note":  types.StringType,
+							},
+						},
+					},
```
dglsparsons, line 237: Potential improvement, add some validation that the list of IPs should have at least 1 element in it. It fails at apply time, which isn't the end of the world though.

vercel/resource_project.go
```diff
@@ -206,11 +210,48 @@ At this time you cannot use a Vercel Project resource with in-line ` + "`environ
...
+				Optional:    true,
+				Attributes: map[string]schema.Attribute{
+					"addresses": schema.SetAttribute{
+						Description:   "The allowed IP addressses and CIDR ranges with optional descriptions.",
+						Required:      true,
+						PlanModifiers: []planmodifier.Set{setplanmodifier.UseStateForUnknown()},
+						ElementType: types.ObjectType{
+							AttrTypes: map[string]attr.Type{
+								"value": types.StringType,
+								"note":  types.StringType,
```
dglsparsons, line 234: This seems to make `note` required, but it's not required in the UI 🤔. Probably fine though.

vercel/resource_project.go
```diff
@@ -206,11 +210,48 @@ At this time you cannot use a Vercel Project resource with in-line ` + "`environ
...
+			"trusted_ips": schema.SingleNestedAttribute{
+				Description: "Ensures only visitors from an allowed IP address can access your deployment.",
+				Optional:    true,
+				Attributes: map[string]schema.Attribute{
+					"addresses": schema.SetAttribute{
+						Description:   "The allowed IP addressses and CIDR ranges with optional descriptions.",
+						Required:      true,
+						PlanModifiers: []planmodifier.Set{setplanmodifier.UseStateForUnknown()},
+						ElementType: types.ObjectType{
+							AttrTypes: map[string]attr.Type{
```
dglsparsons, line 232: I can't seem to remove addresses. Could be how the API is consumed?

dglsparsons: edit: never mind, it's fine. 😖

dglsparsons: edit edit: there is something weird going on here. 
If I add an `address` in the UI, `terraform plan` suggests it will be deleted, but `terraform apply` makes no changes.

vercel/resource_project.go
kitfoster: I updated it, and also marked `value` as sensitive

README.md
```diff
@@ -59,6 +61,22 @@ To run a specific set of tests, use the `-run` flag and specify a regex pattern
...
+$ task install
+```
+
+Build the provider
+
+```sh
+$ task build
+```
+
+Create a `main.tf` file on your machine and use the [terraform cli](https://developer.hashicorp.com/terraform/tutorials/aws-get-started/install-cli#install-terraform) to test
```
dglsparsons, line 78: This is nice, but maybe remove/clean up the section that's already there? 
Screenshot 2023-11-23 at 15 33 00

vercel/resource_project.go
kitfoster: Good idea, updated it 👍

dglsparsons (approved):

vercel/resource_project.go
kitfoster: It worked as expected for me. I added an IP in the UI, then ran terraform apply without the value and it was deleted

README.md
kitfoster: oops I hadn't read that fully, I'll combine them

vercel/resource_project.go
dglsparsons: Interesting.

dglsparsons: Right, yeah, it's my fault. I have `-refresh=false` set in the alias I usually run `terraform apply` with.

dglsparsons (approved):

PR #155, opened 2024-03-08
Fix issue updating sensitive project environment variables

### LINKED ISSUE #154: Cannot update sensitive environment variables
When you try to change the value of a sensitive environment variable, it fails.

image

The provider should do a replacement if the underlying API doesn't allow updating.

We have to be able to update the variable, how else do you rotate secrets?

Workaround is to manually delete the secret in the Vercel dashboard before planning/applying.

dglsparsons:
Closes #154

### DIFF
--- a/client/environment_variable_update.go
+++ b/client/environment_variable_update.go
@@ -10,7 +10,6 @@ import (
 // UpdateEnvironmentVariableRequest defines the information that needs to be passed to Vercel in order to
 // update an environment variable.
 type UpdateEnvironmentVariableRequest struct {
-	Key       string   `json:"key"`
 	Value     string   `json:"value"`
 	Target    []string `json:"target"`
 	GitBranch *string  `json:"gitBranch,omitempty"`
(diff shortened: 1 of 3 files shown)


vercel/resource_project_environment_variable_test.go
```diff
@@ -214,7 +214,7 @@ resource "vercel_project_environment_variable" "example_git_branch" {
 resource "vercel_project_environment_variable" "example_sensitive" {
 	project_id = vercel_project.example.id
 	%[3]s
-	key        = "foo_sensitive_updated"
+	key        = "foo_sensitive"
```
dglsparsons, line 217: The test was faulty - updating the key forced a replacement, so it wasn't really testing that the other fields can be updated.

client/environment_variable_update.go
```diff
@@ -10,7 +10,6 @@ import (
 // UpdateEnvironmentVariableRequest defines the information that needs to be passed to Vercel in order to
 // update an environment variable.
 type UpdateEnvironmentVariableRequest struct {
-	Key       string   `json:"key"`
```
dglsparsons, line 13: Key is required-replace, so should not be present in the update request.

craigandrews (dismissed):

craigandrews (approved):

PR #178, opened 2024-05-01
Set User Agent & a bunch of small fixes

dglsparsons:
Set a User-Agent for the Terraform Provider so we can know in our
infrastructure where these requests are originating from.
The user agent is set to `terraform-provider-vercel`, followed by a `/`,
and then the most recently tagged version.

To do this, we use `ldflags` in the `goreleaser` config. This allows a
hardcoded version (in client/request.go) to be replaced. I have updated
the local build of the provider to also use goreleaser.

I've also fixed the tfproviderlint linter.


Schniz (approved):
<picture data-single-emoji=":letsgooooo:" title=":letsgooooo:">:letsgooooo:</picture>

PR #189, opened 2024-06-18
docs: update `ref` description in `resource_deployment` to fix reference to `files`

austin-tildei:
the `ref` description incorrectly says that `ref` is required if `ref` is not specified, when really I believe it means `files`.


dglsparsons:
Good catch! Thank you.

PR #190, opened 2024-06-24
Add Storage Resources [KV, Postgres, Blob]

versecafe:
## Preview
This is still early in development and people are welcome to help, due note the endpoints being used are not documented but are on a versioned endpoint so should be relatively stable, If anyone on the Vercel team could give guarantees that the versioning will be respected that would be great. also if anyone has any issue with how I'm writing this do let me know I have never written go before so no guarantees it'll come out perfect first try.

### DIFF
--- a/client/redis.go
+++ b/client/redis.go
@@ -1,0 +1,160 @@
+package client
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+)
+
+// DONE: convert to redis from edge config
+type RedisInstance struct {
+	Slug          string   `json:"endpoint"`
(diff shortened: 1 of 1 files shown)


yahel2410:
Hi, can we please move this forward?

versecafe:
@yahel2410 Sure I can try to wrap it up soon, is there any specific part that you need?

dglsparsons:
Hi, just to be clear - one of the reasons I've never worked on the storage resources is that they're fundamentally changing a bit in the near future. 

See https://vercel.com/blog/introducing-the-vercel-marketplace for more details. 

We'll explore how to add Marketplace support to the terraform provider, as well as Blob support. But I don't think KV and Posgres will get supported!

PR #196, opened 2024-07-29
Implement vercel_edge_config_item

LukeSheard:
Add a resource for managing individual `edge_config_item` resources. Each resource represents an individual `(key, value)` pair for an Edge Config.

```
resource "vercel_edge_config" "example" {
  name = "example"
}

  name = "edge-config-example"
}

resource "vercel_edge_config_item" "example": {
  edge_config_id = vercel_edge_config.example.id
  key            = "key"
  value          = "value"
}
```

I've essentially used the `upsert` API from the [documentation](https://vercel.com/docs/storage/edge-config/vercel-api#update-your-edge-config-items) for managing items.


vercel/resource_edge_config_item.go
```diff
@@ -0,0 +1,260 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/vercel/terraform-provider-vercel/client"
+)
+
+// Ensure the implementation satisfies the expected interfaces.
+var (
+	_ resource.Resource              = &edgeConfigItemResource{}
+	_ resource.ResourceWithConfigure = &edgeConfigItemResource{}
```
dglsparsons, line 19: Is there a reason not to support Import, too?

vercel/resource_edge_config_item.go
```diff
@@ -0,0 +1,260 @@
...
+				PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplaceIfConfigured(), stringplanmodifier.UseStateForUnknown()},
+			},
+			"key": schema.StringAttribute{
+				Description:   "The ID of the Edge Config store.",
+				Required:      true,
+				PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplace()},
+			},
+			"value": schema.StringAttribute{
+				Description: "The value you want to assign to the key.",
+				Required:    true,
```
dglsparsons, line 77: RequiresReplace too.

vercel/resource_edge_config_item.go
```diff
@@ -0,0 +1,260 @@
...
+		},
+	}
+}
+
+type EdgeConfigItem struct {
+	Key          types.String `tfsdk:"key"`
+	Value        types.String `tfsdk:"value"`
+	TeamID       types.String `tfsdk:"team_id"`
+	EdgeConfigID types.String `tfsdk:"edge_config_id"`
+	Token        types.String `tfsdk:"token"`
```
dglsparsons, line 88: Token isn't defined in the schema above.

vercel/resource_edge_config_item.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+func responseToEdgeConfigItem(out client.EdgeConfigItem) EdgeConfigItem {
+	return EdgeConfigItem{
+		Key:   types.StringValue(out.Key),
+		Value: types.StringValue(out.Value),
```
dglsparsons, line 94: This is missing a lot of fields.

vercel/resource_edge_config_item.go
```diff
@@ -0,0 +1,260 @@
...
+		return
+	}
+
+	out, err := r.client.GetEdgeConfigItem(ctx, client.EdgeConfigItemRequest{
+		Token:        state.Token.ValueString(),
+		TeamID:       state.TeamID.ValueString(),
+		EdgeConfigID: state.EdgeConfigID.ValueString(),
+		Key:          state.Key.ValueString(),
+		Value:        state.Value.ValueString(),
+	})
```
dglsparsons, line 153: that's a lot of fields for a Get request 🤔

vercel/resource_edge_config_item.go
```diff
@@ -0,0 +1,260 @@
...
+
+	diags = resp.State.Set(ctx, result)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+}
+
+// Update is the same as Create
+func (r *edgeConfigItemResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
```
dglsparsons, line 185: We should remove the contents of the Update method, making it `panic` instead. And rely on the schema having `RequiresReplace` for every field.

vercel/resource_edge_config_item.go
LukeSheard: Sorry I had to rush out and forgot to come back and add a comment in my PR about this.

I think I need to write a function which has 4 parts for importing. If the edge config is scoped to the a team then the import ID would need `${edgeConfigId}-${teamId}-${key}-${value}`, unless I'm missing something?

vercel/resource_edge_config_item.go
LukeSheard: I misinterpreted the documentation originally here, and assumed that we'd need to create a token to use the resource but it seems that's not the case.

vercel/resource_edge_config_item.go
LukeSheard: Just double checking here, but the minimum set needed here is `EdgeConfigID, TeamID, Key`. Is it okay to not pass `Value` here?

vercel/resource_edge_config_item.go
dglsparsons: I think it'd need three, i'm not sure the reason for having `value` in the import, given it can read that from the API. 

Should just be `{teamId}/{edgeConfigId}/{key}`

dglsparsons: We have this pattern in other resources 🙏

vercel/resource_edge_config_item.go
```diff
@@ -0,0 +1,260 @@
...
+
+// Ensure the implementation satisfies the expected interfaces.
+var (
+	_ resource.Resource              = &edgeConfigItemResource{}
+	_ resource.ResourceWithConfigure = &edgeConfigItemResource{}
+)
+
+type edgeConfigItemResource struct {
+	client *client.Client
+}
```
dglsparsons, line 25: This needs wiring up in the `provider.go` file. We could also do with tests, and given there's a fairly sane `Read` method I think we could add a data source too.

LukeSheard: I've wired this up and added an equivalent data source.

dglsparsons:
Hi @LukeSheard, is there a reason you closed this? Apologies for not getting back to you sooner on this too - I've been off on Paternity leave for a while. I can take over this PR though, and add support for this.

LukeSheard:
@dglsparsons I think it was by accident to be honest!

LukeSheard:
@dglsparsons What is missing at this point to get this merged? 😄

dglsparsons:
Just tests.

dglsparsons:
FYI @LukeSheard i'll raise a PR in the morning with tests added. Code is complete and it seems to work like a charm with a couple of minor fixes. 🙌

LukeSheard:
@dglsparsons Thanks so much, I haven't had a chance to test the provider locally directly to validate that it works properly. Very appreciative of your support 🙇

dglsparsons:
Closed in favour of #210

PR #203, opened 2024-09-16
[projects] add support for default project max duration

ecklf:


smaeda-ks:
We can probably close this PR in lieu of https://github.com/vercel/terraform-provider-vercel/pull/206

PR #210, opened 2024-10-11
Add support for edge_config_item

dglsparsons:
Supercedes #196 with additional tests + fixes.


client/edge_config_item.go
```diff
@@ -0,0 +1,135 @@
...
+		},
+	))
+
+	tflog.Info(ctx, "deleting edge config token", map[string]interface{}{
+		"url":     url,
+		"payload": payload,
+	})
+	return c.doRequest(clientRequest{
+		ctx:    ctx,
+		method: "PATCH",
```
ecklf, line 102: why is this patch instead of delete?

client/edge_config_item.go
```diff
@@ -0,0 +1,135 @@
...
+			},
+		},
+	))
+	tflog.Info(ctx, "creating edge config token", map[string]interface{}{
+		"url":     url,
+		"payload": payload,
+	})
+	err = c.doRequest(clientRequest{
+		ctx:    ctx,
+		method: "PATCH",
```
ecklf, line 56: why is this patch instead of post?

client/edge_config_item.go
dglsparsons: The endpoint is to update the contents of an edge-config, rather than to manage a single edge config item. So you submit a list of 'operations', saying what you want to happen to each item.

ecklf (approved):

PR #272, opened 2025-02-20
[projects] Support setting `enableAffectedProjectsDeployments`

matthewbinshtok:
As part of [#project-api-on-vercel](https://vercel.slack.com/archives/C088GDPK35Y), we have the need to support setting `enableAffectedProjectsDeployments` when creating/updating Vercel Projects using the Terraform provider. 

This PR:
- adds the field to the relevant structs and schemas for a Vercel Project
- adds the field to `projectCoercedFields` to match the [API endpoint expected behavior](https://github.com/vercel/api/blob/main/services/api-projects/src/endpoints/update-project.ts#L1148)
- adds tests for setting the field explicitly and also for omitting it


vercel/resource_project.go
```diff
@@ -432,6 +432,10 @@ At this time you cannot use a Vercel Project resource with in-line ` + "`environ
 				PlanModifiers: []planmodifier.Bool{boolplanmodifier.UseStateForUnknown()},
 				Description:   "Vercel provides a set of Environment Variables that are automatically populated by the System, such as the URL of the Deployment or the name of the Git branch deployed. To expose them to your Deployments, enable this field",
 			},
+			"enable_affected_projects_deployments": schema.BoolAttribute{
+				Optional:    true,
```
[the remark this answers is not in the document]
dglsparsons: Ah, I see it's omitted if never set, so this works.

vercel/resource_project_test.go
```diff
@@ -833,6 +862,104 @@ resource "vercel_project" "test" {
 `, projectSuffix, teamID)
 }
 
+func testAccProjectConfigWithoutEnableAffectedSet(projectSuffix, teamID string) string {
+	return fmt.Sprintf(`
+resource "vercel_project" "test" {
+  name = "test-acc-project-%s"
+  %s
+  build_command = "npm run build"
```
dglsparsons, line 870: You can remove most of these settings. I think name and the %s for team config are the only required things here. The rest add a lot of noise to the test

dglsparsons (approved):

vercel/resource_project_test.go
matthewbinshtok: done!

PR #275, opened 2025-02-26
Add resource for adding single projects to shared environment variables

jarneson:
This PR adds a new resource for adding individual projects to a given shared environment variable. This is intended to be used with shared environment variables that are not already terraformed via data source, and usage may conflict with shared environment variables already declared by the shared environment variable resource.


vercel/resource_shared_environment_variable_project_link_test.go
```diff
@@ -0,0 +1,97 @@
+package vercel_test
```
matthewbinshtok, line 1: should we test that existing links are not affected?

codybrouwers (approved):
Not super familiar with Go but everything looks correct to me!

docs/resources/shared_environment_variable_project_link.md
```diff
@@ -0,0 +1,44 @@
...
+page_title: "vercel_shared_environment_variable_project_link Resource - terraform-provider-vercel"
+subcategory: ""
+description: |-
+  Links a project to a Shared Environment Variable.
+  ~> This resource is intended to be used alongside a vercel_shared_environment_variable Data Source, not the Resource. The Resource also defines which projects to link to the shared environment variable, and using both vercel_shared_environment_variable and vercel_shared_environment_variable_project_link together results in undefined behavior.
+---
+
+# vercel_shared_environment_variable_project_link (Resource)
+
+Links a project to a Shared Environment Variable.
```
dglsparsons, line 12: Or does it link a shared environment variable to a project? :trollface:

vercel/resource_shared_environment_variable_project_link.go
```diff
@@ -0,0 +1,224 @@
...
+
+func (r *sharedEnvironmentVariableProjectLinkResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
+	var state SharedEnvironmentVariableProjectLink
+	diags := req.State.Get(ctx, &state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	response, err := r.client.GetSharedEnvironmentVariable(ctx, state.TeamID.ValueString(), state.SharedEnvironmentVariableID.ValueString())
```
dglsparsons, line 149: Do we need to handle 404 here, and remove state if the entire shared environment variable is gone?

vercel/resource_shared_environment_variable_project_link.go
```diff
@@ -0,0 +1,224 @@
...
+		return
+	}
+
+	result := SharedEnvironmentVariableProjectLink{
+		TeamID:                      types.StringValue(response.TeamID),
+		SharedEnvironmentVariableID: types.StringValue(response.ID),
+		ProjectID:                   state.ProjectID,
+	}
+
+	for _, pid := range response.ProjectIDs {
```
dglsparsons, line 164: Any reason this can't be a sliced.Contains(response.ProjectIDs, result.ProjectID.ValueString())?

dglsparsons: In fact in fact, should probably use state.ProjectID.ValueString and build result after the check.

vercel/resource_shared_environment_variable_project_link_test.go
```diff
@@ -0,0 +1,97 @@
...
+					resource.TestCheckResourceAttr("vercel_shared_environment_variable_project_link.test", "team_id", testTeam()),
+				),
+			},
+		},
+	})
+}
+
+func testAccSharedEnvironmentVariableProjectLink(name, team string) string {
+	return fmt.Sprintf(`
+data "vercel_endpoint_verification" "test" {
```
dglsparsons, line 75: Not following what this is here for.

dglsparsons (changes requested):

vercel/resource_shared_environment_variable_project_link_test.go
jarneson: i dunno i just copied from some other test probably

vercel/resource_shared_environment_variable_project_link.go
```diff
@@ -0,0 +1,224 @@
...
+		TeamID:                      types.StringValue(response.TeamID),
+		SharedEnvironmentVariableID: types.StringValue(response.ID),
+		ProjectID:                   plan.ProjectID,
+	}
+
+	tflog.Info(ctx, "project unlinked from shared environment", map[string]interface{}{
+		"team_id":                        result.TeamID.ValueString(),
+		"shared_environment_variable_id": result.SharedEnvironmentVariableID.ValueString(),
+		"project_id":                     result.ProjectID.ValueString(),
+	})
```
dglsparsons, line 223: Another thought - should this have an Import, so you can terraform import team_xxx/env_yyy/prj_zzz ?

jarneson: hrmmmmmm.. maybe?

I wouldn't really expect to need to import this, and a creation of an existing resource would just be a no-op. I'll be able to test this as I already have projects linked to shared env vars that I would want to put into terraform.

vercel/resource_shared_environment_variable_project_link.go
jarneson: I don't think so, because if we are Reading it we expect it to exist, and want to detect when it doesn't (e.g. drift) if the shared env var is just gone, then I think that's an error that requires manual intervention rather than trying to recreate the link.

docs/resources/shared_environment_variable_project_link.md
jarneson: i've heard it both ways

vercel/resource_shared_environment_variable_project_link.go
dglsparsons: Ahh, okay, because this has to be used with a data source, right?

dglsparsons (approved):

PR #279, opened 2025-03-05
Fix microfrontends routeObservabilityToThisProject

kitfoster:
The API has been updated to not accept `routeObservabilityToThisProject` for the default app


jarneson (approved):
tests are passing again!

PR #298, opened 2025-04-14
[vercel_team_config] Fix saml dsync for access groups

kitfoster:
At the moment, the `vercel_team_config` resource incorrectly implements Access Group support. If a team has SAML roles mapped access groups, the provider deletes the mappings.

**This is a breaking change**

steps for upgrading 👇 

1. Update schema for `vercel_team_config.saml`
   1. If `vercel_team_config.saml.access_group_id` is specified: remove it
   1. If `vercel_team_config.saml.roles` is specified: change the values of the object from strings to objects with the format `{ role = "<role e.g. VIEWER>" }` | `{ access_group_id = "<access_group_id>" }`
3. Refresh the terraform state with `terraform refresh` or `terraform apply`

```
// before
resource "vercel_team_config" "test_team" {
  ...
  saml = {
    enforced = <boolean>
    access_group_id = "<access_group_id>"
    roles = {
     <directory_group_id> = "<role e.g. VIEWER>"
    }
  }
}
```

```
// after
resource "vercel_team_config" "test_team" {
  ...
  saml = {
    enforced = <boolean>
    roles = {
      <directory_group_id> = {
        role = "<role e.g. VIEWER>"
      }
      <directory_group_id> = {
        access_group_id = "<access_group_id>"
      }
    }
  }
}
```

### DIFF
--- a/client/team.go
+++ b/client/team.go
@@ -19,12 +19,21 @@ type SamlRoleAccessGroupID struct {
 	AccessGroupID string `json:"accessGroupId"`
 }
 
-type SamlRole struct {
+type SamlRoleAPI struct {
 	Role          *string
 	AccessGroupID *SamlRoleAccessGroupID
 }
 
-func (f *SamlRole) UnmarshalJSON(data []byte) error {
+type SamlRolesAPI map[string]SamlRoleAPI
+
+type SamlRole struct {
+	Role          *string `json:"role"`
+	AccessGroupID *string `json:"access_group_id"`
+}
+
+type SamlRoles map[string]SamlRole
+
+func (f *SamlRoleAPI) UnmarshalJSON(data []byte) error {
 	var role string
 	if err := json.Unmarshal(data, &role); err == nil {
 		f.Role = &role
@@ -38,10 +47,8 @@ func (f *SamlRole) UnmarshalJSON(data []byte) error {
 	return fmt.Errorf("received json is neither Role string nor AccessGroupID map")
 }
 
-type SamlRoles map[string]string
-
 func (f *SamlRoles) UnmarshalJSON(data []byte) error {
-	var result map[string]SamlRole
+	var result SamlRolesAPI
 	if err := json.Unmarshal(data, &result); err != nil {
 		return err
 	}
@@ -50,7 +57,14 @@ func (f *SamlRoles) UnmarshalJSON(data []byte) error {
 		k := k
 		v := v
 		if v.Role != nil {
-			tmp[k] = *(v.Role)
+			tmp[k] = SamlRole{
+				Role: v.Role,
+			}
+		}
+		if v.AccessGroupID != nil {
+			tmp[k] = SamlRole{
+				AccessGroupID: &v.AccessGroupID.AccessGroupID,
+			}
 		}
 	}
 	*f = tmp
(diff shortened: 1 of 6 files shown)


vercel/resource_team_config.go
```diff
@@ -125,28 +124,22 @@ func (r *teamConfigResource) Schema(_ context.Context, req resource.SchemaReques
...
-						Validators: []validator.String{
-							stringvalidator.RegexMatches(regexp.MustCompile("^ag_[A-z0-9_ -]+$"), "Access group ID must be a valid access group"),
-							// Validate only this attribute or roles is configured.
-							stringvalidator.ExactlyOneOf(
-								path.MatchRoot("saml.roles"),
-								path.MatchRoot("saml.access_group_id"),
-							),
+						Computed:    true,
+						ElementType: types.ObjectType{
+							AttrTypes: map[string]attr.Type{
```
kitfoster, line 131: I would like to add a validator that exactly one of `role` or `access_group_id` is present, but couldn't make it work

client/team.go
```diff
@@ -120,8 +134,8 @@ func (c *Client) GetTeam(ctx context.Context, idOrSlug string) (t Team, err erro
 }
 
 type UpdateSamlConfig struct {
-	Enforced bool              `json:"enforced"`
-	Roles    map[string]string `json:"roles"`
+	Enforced bool           `json:"enforced"`
+	Roles    map[string]any `json:"roles"`
```
jarneson, line 138: `any` is pretty gnarly here, when we just need to support marshaling, a union type is only a few more lines of code.

vercel/resource_team_config.go
```diff
@@ -222,31 +215,58 @@ type SamlDirectory struct {
...
 	}
-	roles := map[string]string{}
-	if !s.AccessGroupId.IsNull() {
-		roles["accessGroupId"] = s.AccessGroupId.ValueString()
-	} else {
-		s.Roles.ElementsAs(ctx, &roles, false)
+	roles := map[string]any{}
+	for k, v := range s.Roles {
+		if v.Role.IsNull() && v.AccessGroupID.IsNull() {
+			panic("SAML roles must specify either a role or access group id: " + k)
```
kitfoster, line 258: Would like to know how to correctly throw here

client/team.go
jarneson: something like:

```
type UpdatSamlConfigRole struct {
  Role *string
  AccessGroupID *string
}

func (r *UpdateSamlConfigRole) MarshalJSON() ([]byte, error) {
  if r.Role != nil { return json.Marshal(*r.Role) }
  if r.AccessGroupID != nil { return json.Marshal(*r.AccessGroupID) }
  return nil, fmt.Errorf("bad union")
}
```

vercel/resource_team_config.go
jarneson: i think this just ends up being a marshalling error when we go to hit the RPCs? we could explicitly check it, not sure if terraform provides a hook for that. doug knows way more than I do on this stuff.

vercel/resource_team_config.go
```diff
@@ -124,29 +123,39 @@ func (r *teamConfigResource) Schema(_ context.Context, req resource.SchemaReques
...
+								"access_group_id": schema.StringAttribute{
+									Description: "The access group id to assign to the user.",
+									Optional:    true,
+								},
+							},
+							// Not sure why, but this does not work
+							// Validators: []validator.Object{
+							// 	objectvalidator.ExactlyOneOf(
+							// 		path.MatchRelative().AtName("role"),
+							// 		path.MatchRelative().AtName("access_group_id"),
```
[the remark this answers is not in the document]
kitfoster: Added a custom validator

dglsparsons (approved):

kitfoster:
The tests currently have a failure, as the schema upgrade from 0 to 1 means that `terraform apply` performs a refresh. This shows that the resource will be updated in place. The upgrade steps show that a `terraform apply` or `terraform refresh` is required for the new version, so this is an expected failure

PR #313, opened 2025-05-02
[firewall config resource] Add bot filter support

OhMyVolk:
Add terraform support for `bot_filter` managed ruleset.

API already supports this.
Tested locally with a project


dglsparsons (approved):

OhMyVolk:
Getting failure in tests
```
 Trying to access resource under scope "vercel-terraform-test-ci". You must
        re-authenticate to this scope or use a token with access to this scope.
```
I think we need to refresh some access token
@dglsparsons

ctgowrie (commented):
LGTM

PR #316, opened 2025-05-06
fix(go): change package to v3

dglsparsons:


ecklf (approved):

omercnet:
thanks @dglsparsons @ecklf 
please also issue a releast/tag so it can be consumed with @latest and not @main

PR #319, opened 2025-05-09
[enhanced builds] add build_machine_type option to project creation

marc-vercel:
After the release of on-demand enhanced builds, we are adding support for this feature at the project level. We are adding a new option, `build_machine_type`, to the project.

Requisites of this feature:
- The project option `on_demand_concurrent_builds` must be set to true. If it is set to false and `build_machine_type` is set to something, we should error.
- It is optional (no need to set it, even if `on_demand_concurrent_builds` is set to true).
- It can have two values: `standard` and `enhanced`.

## CURL Examples

Setting the `build_machine_type` to enhanced:

```
curl --request PATCH \
  --url https://api.vercel.com/v9/projects/<project_id>\
  --header 'Authorization: Bearer <token>' \
  --header 'Content-Type: application/json' \
  --data '{
  "resourceConfig": {
    "elasticConcurrencyEnabled": true,
    "buildMachineType": "enhanced"
  }
}'
```

Setting the `build_machine_type` to standard (notice that we send null instead of standard):

```
curl --request PATCH \
  --url https://api.vercel.com/v9/projects/<project_id>\
  --header 'Authorization: Bearer <token>' \
  --header 'Content-Type: application/json' \
  --data '{
  "resourceConfig": {
    "elasticConcurrencyEnabled": true,
    "buildMachineType": null
  }
}'
```

### DIFF
--- a/client/project.go
+++ b/client/project.go
@@ -225,6 +225,7 @@ type ResourceConfigResponse struct {
 	FunctionDefaultTimeout    *int64  `json:"functionDefaultTimeout"`
 	Fluid                     bool    `json:"fluid"`
 	ElasticConcurrencyEnabled bool    `json:"elasticConcurrencyEnabled"`
+	BuildMachineType          *string `json:"buildMachineType"`
 }
 
 type ResourceConfig struct {
@@ -232,6 +233,7 @@ type ResourceConfig struct {
 	FunctionDefaultTimeout    *int64  `json:"functionDefaultTimeout,omitempty"`
 	Fluid                     *bool   `json:"fluid,omitempty"`
 	ElasticConcurrencyEnabled *bool   `json:"elasticConcurrencyEnabled,omitempty"`
+	BuildMachineType          *string `json:"buildMachineType,omitempty"`
 }
 
 // GetProject retrieves information about an existing project from Vercel.
--- a/vercel/data_source_project.go
+++ b/vercel/data_source_project.go
@@ -393,6 +393,14 @@ For more detailed information, please see the [Vercel documentation](https://ver
 				Optional:    true,
 				Computed:    true,
 			},
+			"build_machine_type": schema.StringAttribute{
+				Description: "When `on_demand_concurrent_builds` is enabled, choose the build machine: `standard` (4 vCPU, 8 GB) or `enhanced` (8 vCPU, 16GB)",
+				Optional:    true,
+				Computed:    true,
+				Validators: []validator.String{
+					stringvalidator.OneOf("standard", "enhanced"),
+				},
+			},
 		},
 	}
 }
@@ -437,6 +445,7 @@ type ProjectDataSource struct {
 	ResourceConfig                      types.Object          `tfsdk:"resource_config"`
 	NodeVersion                         types.String          `tfsdk:"node_version"`
 	OnDemandConcurrentBuilds            types.Bool            `tfsdk:"on_demand_concurrent_builds"`
+	BuildMachineType                    types.String          `tfsdk:"build_machine_type"`
 }
 
 func convertResponseToProjectDataSource(ctx context.Context, response client.ProjectResponse, plan Project, environmentVariables []client.EnvironmentVariable) (ProjectDataSource, error) {
@@ -503,6 +512,7 @@ func convertResponseToProjectDataSource(ctx context.Context, response client.Pro
 		ResourceConfig:                      project.ResourceConfig,
 		NodeVersion:                         project.NodeVersion,
 		OnDemandConcurrentBuilds:            project.OnDemandConcurrentBuilds,
+		BuildMachineType:                    project.BuildMachineType,
 	}, nil
 }
 
--- a/vercel/resource_project.go
+++ b/vercel/resource_project.go
@@ -579,6 +579,15 @@ At this time you cannot use a Vercel Project resource with in-line ` + "`environ
 				Computed:      true,
 				PlanModifiers: []planmodifier.Bool{boolplanmodifier.UseStateForUnknown()},
 			},
+			"build_machine_type": schema.StringAttribute{
+				Description: "When `on_demand_concurrent_builds` is enabled, choose the build machine: `standard` (4 vCPU, 8 GB) or `enhanced` (8 vCPU, 16GB)",
+				Optional:    true,
+				Computed:    true,
+				Validators: []validator.String{
+					stringvalidator.OneOf("standard", "enhanced"),
+				},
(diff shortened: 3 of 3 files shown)


client/project.go
```diff
@@ -225,13 +225,15 @@ type ResourceConfigResponse struct {
...
 	ElasticConcurrencyEnabled bool    `json:"elasticConcurrencyEnabled"`
+	BuildMachineType          *string `json:"buildMachineType"`
 }
 
 type ResourceConfig struct {
 	FunctionDefaultMemoryType *string `json:"functionDefaultMemoryType,omitempty"`
 	FunctionDefaultTimeout    *int64  `json:"functionDefaultTimeout,omitempty"`
 	Fluid                     *bool   `json:"fluid,omitempty"`
 	ElasticConcurrencyEnabled *bool   `json:"elasticConcurrencyEnabled,omitempty"`
+	BuildMachineType          *string `json:"buildMachineType,omitempty"`
```
dglsparsons, line 236: `omitempty` here means we'll never actually send `null` to the API. Probably want to remove that.

vercel/data_source_project.go
```diff
@@ -393,6 +393,14 @@ For more detailed information, please see the [Vercel documentation](https://ver
 				Optional:    true,
 				Computed:    true,
 			},
+			"build_machine_type": schema.StringAttribute{
+				Description: "When `on_demand_concurrent_builds` is enabled, choose the build machine: `standard` (4 vCPU, 8 GB) or `enhanced` (8 vCPU, 16GB)",
+				Optional:    true,
+				Computed:    true,
```
dglsparsons, line 399: I think we can add `UseStateForUnknown`, as this setting will never change based on the other settings changing, right? 
Or does disabling `on_demand_concurrent_builds` affect this?

dglsparsons: edit: i'm a fool, this is the data source.

vercel/data_source_project.go
```diff
@@ -393,6 +393,14 @@ For more detailed information, please see the [Vercel documentation](https://ver
 				Optional:    true,
 				Computed:    true,
 			},
+			"build_machine_type": schema.StringAttribute{
+				Description: "When `on_demand_concurrent_builds` is enabled, choose the build machine: `standard` (4 vCPU, 8 GB) or `enhanced` (8 vCPU, 16GB)",
+				Optional:    true,
```
dglsparsons, line 398: Shouldn't be optional on the data source. Only computed. (it's a read-only property here).

vercel/data_source_project.go
```diff
@@ -393,6 +393,14 @@ For more detailed information, please see the [Vercel documentation](https://ver
...
+				Computed:    true,
+				Validators: []validator.String{
```
dglsparsons, line 400: No need for validators on the data source.

vercel/resource_project.go
```diff
@@ -579,6 +579,15 @@ At this time you cannot use a Vercel Project resource with in-line ` + "`environ
...
 				PlanModifiers: []planmodifier.Bool{boolplanmodifier.UseStateForUnknown()},
 			},
+			"build_machine_type": schema.StringAttribute{
+				Description: "When `on_demand_concurrent_builds` is enabled, choose the build machine: `standard` (4 vCPU, 8 GB) or `enhanced` (8 vCPU, 16GB)",
+				Optional:    true,
+				Computed:    true,
+				Validators: []validator.String{
+					stringvalidator.OneOf("standard", "enhanced"),
+				},
+				PlanModifiers: []planmodifier.String{stringplanmodifier.UseStateForUnknown()},
```
dglsparsons, line 589: What happens if `on_demand_concurrent_builds` is disabled?

vercel/resource_project.go
```diff
@@ -579,6 +579,15 @@ At this time you cannot use a Vercel Project resource with in-line ` + "`environ
 				Computed:      true,
 				PlanModifiers: []planmodifier.Bool{boolplanmodifier.UseStateForUnknown()},
 			},
+			"build_machine_type": schema.StringAttribute{
+				Description: "When `on_demand_concurrent_builds` is enabled, choose the build machine: `standard` (4 vCPU, 8 GB) or `enhanced` (8 vCPU, 16GB)",
+				Optional:    true,
+				Computed:    true,
+				Validators: []validator.String{
```
dglsparsons, line 586: Let's figure out how to add a validator to make sure `on_demand_concurrent_builds` is enabled.

dglsparsons: We have `ModifyPlan` on this resource - we can do it there.

vercel/resource_project.go
```diff
@@ -1101,6 +1111,11 @@ func (r *ResourceConfig) toClientResourceConfig(onDemandConcurrentBuilds types.B
 	if !onDemandConcurrentBuilds.IsUnknown() {
 		resourceConfig.ElasticConcurrencyEnabled = onDemandConcurrentBuilds.ValueBoolPointer()
 	}
+	if !buildMachineType.IsUnknown() && *resourceConfig.ElasticConcurrencyEnabled {
```
dglsparsons, line 1114: What if `resourceConfig.ElasticConcurrencyEnabled` is `null`?

dglsparsons: I don't think we need this logic here, we can just use a validator, and set `buildMachineType` without worrying about its relationship to elastic-concurrency.

vercel/resource_project.go
```diff
@@ -1101,6 +1111,11 @@ func (r *ResourceConfig) toClientResourceConfig(onDemandConcurrentBuilds types.B
...
+		// `build_machine_type` can only be set when `on_demand_concurrent_builds`
+		// is enabled.
+		resourceConfig.BuildMachineType = buildMachineType.ValueStringPointer()
```
dglsparsons, line 1117: Don't forget to convert `standard` to a null pointer?

vercel/resource_project.go
```diff
@@ -1472,6 +1487,7 @@ func convertResponseToProject(ctx context.Context, response client.ProjectRespon
 		ResourceConfig:                      resourceConfig,
 		NodeVersion:                         types.StringValue(response.NodeVersion),
 		OnDemandConcurrentBuilds:            types.BoolValue(response.ResourceConfig.ElasticConcurrencyEnabled),
+		BuildMachineType:           		 types.StringValue(*response.ResourceConfig.BuildMachineType),
```
dglsparsons, line 1490: Don't forget to convert `null` to `standard`? Also, dereferencing the pointer he could panic.

PR #325, opened 2025-05-15
Flesh out secure_compute_network data source

dglsparsons:
This will be useful in the future for getting the ID of a network and
being able to associate it with projects.
Plus, there may be use-cases where getting the AWS IDs, or regions, etc.
is useful.


vercel/data_source_secure_compute_network.go
```diff
@@ -0,0 +1,312 @@
...
+					"account_id": schema.StringAttribute{
+						Description: "The AWS account ID.",
+						Computed:    true,
+					},
+					"region": schema.StringAttribute{
+						Description: "The AWS region.",
+						Computed:    true,
+					},
+					"elastic_ip_addresses": schema.SetAttribute{
+						Description: "A list of Elastic IP addresses.",
```
marcgreenstock, line 121:
```suggestion
						Description: "A list of Public IP addresses.",
```

Also, we should move this out of the AWS struct.

vercel/data_source_secure_compute_network.go
```diff
@@ -0,0 +1,312 @@
...
+			},
+			"cidr_block": schema.StringAttribute{
+				Description: "The CIDR block assigned to the Secure Compute Network.",
+				Computed:    true,
+			},
+			"availability_zone_ids": schema.SetAttribute{
+				Description: "A set of AWS Availability Zone IDs where the Secure Compute Network resources are deployed.",
+				Computed:    true,
+				ElementType: types.StringType,
+			},
```
marcgreenstock, line 99: This should be in the AWS struct.

PR #336, opened 2025-06-03
rolling release config

dimitropoulos:
This PR adds controls for the rolling release config via terraform.

PR #338, opened 2025-06-05
[tf] rolling release configuration support

brookemosby:
Picking up from dimitri's PR
https://github.com/vercel/terraform-provider-vercel/pull/336


dimitropoulos:
thanks! pulled into #336

PR #344, opened 2025-06-11
Create dsync groups data source

marcgreenstock:
Adds the `vercel_dsync_groups` data source:

```tf
data "vercel_dsync_groups" "acme" {
  team_id = "team_xyz"
}
```


docs/data-sources/dsync_groups.md
```diff
@@ -0,0 +1,32 @@
...
+subcategory: ""
+description: |-
+  Provides information about DSync groups for a team.
+---
+
+# vercel_dsync_groups (Data Source)
+
+Provides information about DSync groups for a team.
+
+
```
dglsparsons, line 13: Can you add an example in the examples directory?

vercel/data_source_dsync_groups.go
```diff
@@ -0,0 +1,132 @@
...
+	d.client = client
+}
+
+func (d *dsyncGroupsDataSource) Schema(_ context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Provides information about DSync groups for a team.",
+		Attributes: map[string]schema.Attribute{
+			"team_id": schema.StringAttribute{
+				Description: "The ID of the team to retrieve DSync groups for.",
+				Required:    true,
```
dglsparsons, line 54: This shouldn't be required. We use the team_id configured on the provider itself if not present. e.g.
```
			"team_id": schema.StringAttribute{
				Optional:      true,
				Computed:      true,
				PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplaceIfConfigured(), stringplanmodifier.UseStateForUnknown()},
				Description:   "The team ID to add the project to. Required when configuring a team resource if a default team has not been set in the provider.",
			},
```

vercel/data_source_dsync_groups.go
```diff
@@ -0,0 +1,132 @@
...
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	out, err := d.client.GetDsyncGroups(ctx, config.TeamID.ValueString())
+	if client.NotFound(err) {
+		resp.State.RemoveResource(ctx)
+		return
+	}
```
dglsparsons, line 98: Does this fail to emit `Diagnostics` if we get a NotFound error then? Not sure on the behaviour here, but this doesn't feel right for a data source. 
Maybe just omit this check, as the below error catching should be fine.

dglsparsons (changes requested):

vercel/data_source_dsync_groups.go
dglsparsons: e.g. this should be fine to do: 
```
provider "vercel" {
   team = "team_xxxx" 
}

data "vercel_dsync_groups" "example" {}
```

dglsparsons (approved):

PR #360, opened 2025-07-14
Add support for previewDeploymentsDisabled in project

blurrah:
This adds the `previewDeploymentsDisabled` boolean flag which can be used to completely disable preview deployments for a project.


blurrah:
Closing due to this field only being available on project update (PATCH) not on creation.

PR #397, opened 2025-09-18
Add Vercel Drain resource

lukahartwig:


vercel/data_source_drain.go
```diff
@@ -0,0 +1,387 @@
+package vercel
```
dglsparsons, line 1: needs tests :)

lukahartwig:
Will re-open when I have time to finish it.

PR #409, opened 2025-10-21
fix: Handle existence operators (ex/nex) in firewall rules

SIGTERM-015:
## Fix

The issue was that the code assumed `condition.Value` is always a string, but for existence operators the API returns `null` or empty string (since these operators only check if a field exists, not its value).

Changes:
- Added `isExistenceOp()` helper to identify `ex`/`nex` operators
- Updated `fromCondition()` to handle nil values safely
- Modified `Conditions()` to send empty string to API for existence operators
- Added tests for existence operators

## Usage

Now you can create rules with existence operators without crashes:

```hcl
resource "vercel_firewall_config" "example" {

  rules {
    rule {
      name = "require_auth"
      action = { action = "deny" }
      condition_group = [{
        conditions = [{
          type = "header"
          op   = "nex"
          key  = "Authorization"
          # no 'value' field needed
        }]
      }]
    }
  }
}
```

Tested with real API - creating, reading, and importing configs all work now.

### DIFF
--- a/docs/resources/firewall_config.md
+++ b/docs/resources/firewall_config.md
@@ -78,7 +78,8 @@ resource "vercel_firewall_config" "example" {
             # Note: no 'value' field needed for existence operators (ex/nex)
             type = "cookie"
             key  = "_session"
-            op   = "nex"
+            neg  = true
+            op   = "ex"
         }]
       }]
       action = {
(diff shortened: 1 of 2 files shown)


dglsparsons:
Hey @SIGTERM-015 - thanks for raising this. 

The changes LGTM, I've just asked our firewall team to verify them, too, so i'll be back as soon as they've had a look. 

Really appreciate these contributions.

SIGTERM-015:
And when will that be? 
I have no intend to rush but I'm using this changes in prod and my team is having to use dev overrides pointing to the binaries I have compiled. 
It's not a big deal but the sooner I get rid of that the better.


Thanks for the quick merge btw <3

PR #413, opened 2025-11-18
Fix log-drain resource by using new API endpoints

### LINKED ISSUE #412: [BUG] Resource "vercel_log_drain" does not support "redirect" source, and "x-vercel-verify" is not required by Sentry
### Describe the bug

1. Resource creation fails due to the x-vercel-verify header no longer being required by Sentry
2. "redirect" `source` type is not supported but a valid value in Vercel

Workaround of creating manually in Vercel & then importing into TF for #1 only works if you never try to update the resource, or it will delete and fail to re-create.

No workaround for #2 either

### To Reproduce

```
resource "vercel_log_drain" "dashboards_to_sentry" {
  team_id = "REDACTED"

  delivery_format = "json"
  endpoint        = var.sentry_dsn

  headers = {
    
    "x-sentry-auth" = "sentry sentry_key=${var.sentry_key}"
    "Content-Type"  = "application/json"

  }

  environments = ["production"]
  # sources      = ["lambda", "redirect", "firewall"]
  sources      = ["lambda", "firewall"]

  project_ids = local.dashboard_project_ids
}
```

### Terraform and Vercel Provider Version

```console
Terraform v1.13.5
provider registry.opentofu.org/vercel/vercel v4.0.0
```

### Additional Details

This may not be everything - just what I have run into.

dglsparsons:
Closes #412

### DIFF
--- a/client/log_drain.go
+++ b/client/log_drain.go
@@ -7,9 +7,12 @@ import (
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 )
 
+// LogDrain represents a simplified drain shape used by the provider
+// regardless of the underlying API representation.
 type LogDrain struct {
 	ID             string            `json:"id"`
 	TeamID         string            `json:"ownerId"`
+	Name           string            `json:"name"`
 	DeliveryFormat string            `json:"deliveryFormat"`
 	Environments   []string          `json:"environments"`
 	Headers        map[string]string `json:"headers"`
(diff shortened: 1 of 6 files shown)


client/log_drain.go
```diff
@@ -32,31 +38,197 @@ type CreateLogDrainRequest struct {
...
+
+type drainsDeploymentFilter struct {
+	Environments []string `json:"environments,omitempty"`
+}
+
+type drainsDeliveryHTTP struct {
+	Type        string            `json:"type"` // "http"
+	Endpoint    string            `json:"endpoint"`
+	Encoding    string            `json:"encoding"`              // json | ndjson
+	Compression string            `json:"compression,omitempty"` // gzip | none
```
kakadiadarpan: you can drop `compression` as it isn't supported yet

Issue #354, opened 2025-06-23
✨ Make environment variables write only


holbrookab-bond:
### What functionality would you like to see?

Hi @dglsparsons , per the other issue, I started playing around with the idea of making the environment variables write only in terraform side, to limit the state writing of the values in the code.  This would require a plugin upgrade.

Feel free to take a look at:
https://github.com/holbrookab-bond/terraform-provider-vercel/pull/1

Implementation details:
* Terraform doesn't support WriteOnly in Sets, so I made this a Map (to avoid sorting issues).  Note: this would break the existing provider.
* If an environment variable's value is updated, the *entire* resource must be destroyed.  Per my research it doesn't appear you can just do an update.
* I started touching the regular environment variable, but decided to focus on the main environment variables plural one.
* I haven't looked at other resources that may need updates like this.

This is *not* ready for a PR for you (I've made no real changes to documentation or tests), but thought you'd like to see and react to it.

Thanks!

holbrookab-bond:
I mean, Terraform supports "lists", just not sets.   That just means you need to support the state management I guess of sorting the list.

> it's perfectly valid for the same environment key to appear multiple times (e.g. once per environment).
Testing on my test Vercel project, though, you could just have multiple environment variable resources linked to the same project.  It handles it fine, I think because each environment variable addition will get a unique ID tied to it?

<img width="934" alt="Image" src="https://github.com/user-attachments/assets/8e33e345-dccc-4174-910b-19ea15bf38e8" />

Came from
```
# Example: Add environment variables to test your drift detection
resource "vercel_project_environment_variables" "test" {
  project_id = data.vercel_project.my_repo.id
  variables = {
    "TEST_VAR" = {
      value  = "hello"
      target = ["production", "preview"]
    },
    "SECRET_VAR" = {
      value     = "my super secret"
      sensitive = true
      target    = ["production"]
    }
  }
}

resource "vercel_project_environment_variables" "test2" {
  project_id = data.vercel_project.my_repo.id
  variables = {
    "TEST_VAR" = {
      value  = "world"
      target = ["development"]
    },
    "ANOTHER_TEST" = {
      value  = "hello again"
      target = ["development"]
    },
    "SECRET_VAR" = {
      value     = "secret_valuez"
      sensitive = true
      target    = ["preview"]
    }
  }
}
```

dglsparsons:
haha, yeah, that's fair. A very weird caveat to work around though, and hard to document.

[...]

PR #417, opened 2025-11-29
feat: write-only values support for resource.project_environment_variable

corey-livepeer:
This is an implementation of https://github.com/vercel/terraform-provider-vercel/issues/354

Because Terraform does not allow nested write-only values in sets, that would lead to a breaking change of `resource.project_environment_variables` to convert it from sets to maps.

Instead, I simply added a new `values_wo` field ([as recommended by Terraform](https://developer.hashicorp.com/terraform/language/manage-sensitive-data/ephemeral#write-only-arguments), see [aws_ssm_parameter](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ssm_parameter#value_wo-1) as an example) to `resource.project_environment_variable`, and validation checks to ensure only one of `value` or `value_wo` is used.

Docs were updated by running `task docs`


corey-livepeer:
@dglsparsons is this a good start to adding write-only support?

kokolight:
can we prioritize this? this is very important to make a secure variable creation in Vercel

PR #443, opened 2026-01-15
fix: add computed id attribute to team_config and edge_config_schema resources

omercnet:
This pull request updates the schema for resource IDs in the Vercel provider to ensure that IDs are treated as computed outputs rather than required inputs. This change brings consistency to how resource IDs are handled and aligns with best practices for Terraform providers.

**Schema attribute updates:**

* Changed the `id` attribute in `resource_edge_config_schema.go` to be `Computed: true` instead of `Required: true`, ensuring it is treated as an output rather than an input.
* Changed the `id` attribute in `resource_team_config.go` to be `Computed: true` and removed the `UseNonNullStateForUnknown` plan modifier, making the attribute a computed output.

**Test logic update:**

* Updated the `hasIDAttribute` function in `resource_id_test.go` to only return true if the `id` attribute is computed, not required, reflecting the new schema logic.


dglsparsons:
Hi @omercnet. I'm not following this one: 

> Fixed schema validation error in vercel_team_config resource where the id attribute was incorrectly marked as both Required and Computed, causing Terraform to reject the schema.

It's not marked as both at the moment, only Required, no? 

And this is correct - you need to pass the edge_config_id to the edge_config_schema, in order to tell it which edge_config the schema is for. 

Ditto for the team_config resource. You need to tell it which team this config is for.

omercnet:
@dglsparsons sorry for the confusing - fixed this pr now and it's ready for review

omercnet:
answering myself: yes - the client requires it (the root cause is the REST API design), which is silly because it's derived... making it optional with a check to ensure it's provided to ensure terraform/pulumi compatibility while maintaining compatibility with the API

dglsparsons:
> would you rather we fix the go client ?

I don't follow what you're proposing here.

dglsparsons:
Closing this out as stale. The proposed change did not reach agreement: these fields are still required to identify what the resource operates on, and making them optional would move the failure from plan/schema validation to apply time. The PR has also been inactive for several months and the checks are outdated.

If you want to revisit Pulumi compatibility here, please open a fresh issue or PR with the specific bridge failure and the desired Terraform schema contract.

PR #470, opened 2026-04-14
Terraform resource schema and client calls for secure compute

matthewbinshtok:


simoncroak:
@matthewbinshtok any chance we can convert this into a 'real' PR (not a draft) I would love to get the feature merged. Thanks!

PR #472, opened 2026-04-17
Add vercel_project_protection_bypass resource

kitfoster:
## Summary
- Adds a new `vercel_project_protection_bypass` resource for managing multiple deployment protection automation bypasses per project, matching the current API surface (multiple named bypasses, each with a `note` and an `is_env_var` flag — the latter picks which bypass is exposed as `VERCEL_AUTOMATION_BYPASS_SECRET` on deployments).
- New client helpers: `CreateProtectionBypass`, `UpdateProtectionBypass`, `DeleteProtectionBypass`, `DemoteProtectionBypass`, `GetProtectionBypass`. Extended the `ProtectionBypass` struct with `IsEnvVar`, `Note`, `CreatedAt`, `CreatedBy`, `IntegrationID`, `ConfigurationID`.
- On create, if the user requested `is_env_var = true` but the API didn't make the new bypass the default (i.e. another one still held that slot), the provider issues a follow-up `update` to promote it — the API atomically demotes the previous default.
- Per-project mutex (`sync.Map` keyed by project ID) serialises all bypass mutations so the generate-then-diff identification and pre-delete promotion paths can't race each other.
- Payloads are redacted before being logged via tflog (any `secret` key at any nesting depth).

## ⚠️ Breaking changes

1. **`vercel_project.protection_bypass_for_automation` and `protection_bypass_for_automation_secret` are removed from the resource schema.** Configs that set either attribute will fail with an unknown-attribute error. Users must migrate to the new `vercel_project_protection_bypass` resource and `terraform state rm` / `import` to bring existing bypasses under the new resource.
2. **`data.vercel_project.protection_bypass_for_automation_secret` changes type** from `String` to `List(String)` (now returns every automation-bypass secret on the project). Downstream references that consume it as a string will need to index into the list. The attribute is marked `Deprecated`; prefer the new resource going forward.

Both changes need a CHANGELOG entry and a migration note in the release.

## Example

```hcl
  name = "example-project"
}

resource "vercel_project_protection_bypass" "ci" {
  note       = "ci pipeline"
  # first bypass is automatically the env-var default
}

resource "vercel_project_protection_bypass" "qa" {
  note       = "preview QA"
  secret     = var.qa_bypass_secret  # optional; else Vercel generates one
}
```

## Test plan
- [x] `go build ./... && go vet ./...`
- [x] `go test ./client/... ./vercel/...` (non-acceptance)
- [ ] `TF_ACC=1 go test ./vercel -run TestAcc_ProjectProtectionBypass -v` against a team with the `EnableMultipleAutomationBypasses` LaunchDarkly flag enabled
- [ ] Manual upgrade test: take a workspace using the old `protection_bypass_for_automation` attributes, upgrade the provider, confirm the error is clear and the documented migration (`state rm` + `import`) works end-to-end
- [ ] Manual `terraform apply` on `examples/resources/vercel_project_protection_bypass/` — verify create/swap/import/destroy end-to-end

## Notes
- Only the `automation-bypass` scope is managed. `integration-automation-bypass` entries (created via OAuth integration tokens) are intentionally out of scope.
- `secret` triggers replacement — rotating it means destroy + create. `note` and `is_env_var` update in place.
- Import format: `team_id/project_id/secret` or `project_id/secret`.


client/deployment_protection.go
```diff
@@ -24,7 +24,13 @@ type TrustedIps struct {
 }
 
 type ProtectionBypass struct {
-	Scope string `json:"scope"`
+	Scope           string  `json:"scope"`
+	IsEnvVar        *bool   `json:"isEnvVar,omitempty"`
+	Note            *string `json:"note,omitempty"`
+	CreatedAt       int64   `json:"createdAt,omitempty"`
+	CreatedBy       string  `json:"createdBy,omitempty"`
```
dglsparsons, line 31: I think we could omit these?

dglsparsons: Wait, is this the response or the request 👀

kitfoster: can do 👍

kitfoster: claude says they're used, happy to keep them?

dglsparsons (approved):

PR #480, opened 2026-04-22
fix: filter automation-bypass scope when parsing update response

mehulkar:
## Summary

`UpdateProtectionBypassForAutomation` errored with `the response contained an unexpected number of items (N)` whenever the project had any other bypass entry (e.g. a shareable-link bypass) in its `protectionBypass` map, because the client asserted `len(response.ProtectionBypass) == 1`.

A project's `protectionBypass` is keyed by secret and each entry has a `scope` — the automation-bypass flow only owns entries where `scope == "automation-bypass"`. The Read path already filters this way at `vercel/resource_project.go:1737`. This PR applies the same filter in the update client.

## Changes

- `client/project_protection_bypass_for_automation_update.go`
  - Short-circuit to the caller-supplied `NewSecret` when it's set. We know what we wrote, so we don't need to inspect the response — this also handles the rotation path (generate+revoke) where the API may transiently return both old and new automation-bypass entries.
  - Otherwise filter the response by `scope == "automation-bypass"` and return the single matching key. Error only if the filtered count isn't 1.
- `client/project_protection_bypass_for_automation_update_test.go` (new)
  - Multi-scope response returns the automation-bypass key.
  - Rotation with caller-supplied `NewSecret` returns that secret directly even when the response has two automation-bypass entries.
  - Missing automation-bypass entry produces a clear error.
  - `NewValue=false` (disable) ignores the response map entirely.

## Repro

Reported by a user whose plan produced:
```
Error: Error updating protection bypass for automation
...
unexpected error updating Protection Bypass For Automation: error adding
protection bypass for automation: the response contained an unexpected
number of items (2)
```
The project had an additional non-automation-bypass entry in the `protectionBypass` map, so the strict count check rejected an otherwise-successful API response.

## Test plan

- [x] `go test ./client/...` passes including the 4 new cases.
- [x] `go vet ./...` clean.
- [ ] Verify against the reporter's project (apply with a real secret rotation / fresh generate on a project that already has a shareable-link bypass).

### DIFF
--- a/client/project_protection_bypass_for_automation_update.go
+++ b/client/project_protection_bypass_for_automation_update.go
@@ -98,14 +98,29 @@ func (c *Client) UpdateProtectionBypassForAutomation(ctx context.Context, reques
 		return
 	}
 
-	if len(response.ProtectionBypass) != 1 {
-		return s, fmt.Errorf("error adding protection bypass for automation: the response contained an unexpected number of items (%d)", len(response.ProtectionBypass))
+	// When the caller supplied an explicit new secret we already know what was
+	// set — no need to inspect the response, which can contain entries for
+	// other bypass scopes (e.g. shareable-link) or both the old and new
+	// automation-bypass during a rotation.
+	if request.NewSecret != "" {
+		return request.NewSecret, nil
 	}
 
-	// return the first key from the map
-	for key := range response.ProtectionBypass {
-		return key, err
+	// The API generated a secret for us. Filter to the automation-bypass scope
+	// since the project may have other bypass scopes (e.g. shareable-link) in
+	// the same response map.
+	var automationKey string
+	var automationCount int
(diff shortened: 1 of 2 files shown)


client/project_protection_bypass_for_automation_update.go
```diff
@@ -98,14 +98,29 @@ func (c *Client) UpdateProtectionBypassForAutomation(ctx context.Context, reques
...
+	// automation-bypass during a rotation.
+	if request.NewSecret != "" {
+		return request.NewSecret, nil
 	}
 
-	// return the first key from the map
-	for key := range response.ProtectionBypass {
-		return key, err
+	// The API generated a secret for us. Filter to the automation-bypass scope
+	// since the project may have other bypass scopes (e.g. shareable-link) in
```
kitfoster, line 110: not shareable links, just other automation bypasses

PR #484, opened 2026-04-24
fix(project): allow coexistence of serverless_function_region and resource_config.function_default_regions

### LINKED ISSUE #475: [BUG] `vercel_project` import fails on pre-migration projects due to `ConflictsWith` between `serverless_function_region` and `resource_config.function_default_regions`
### Describe the bug

`vercel_project` resource cannot be imported for projects that have BOTH `serverless_function_region` (legacy) and `resource_config.function_default_regions` (modern) populated in Vercel's stored state.

The provider's schema declares these two fields as `ConflictsWith` each other, so importing a project whose live state contains both fails with two "Invalid Attribute Combination" errors:

```
Error: Invalid Attribute Combination. Attribute
"resource_config.function_default_regions" cannot be specified when
"serverless_function_region" is specified.

Error: Invalid Attribute Combination. Attribute
"serverless_function_region" cannot be specified when
"resource_config.function_default_regions" is specified.
```

The coexistence is normal state on pre-`resource_config`-migration projects that were later touched by the modern config path — Vercel's API keeps the two fields read/write-synced (writing one mirrors to the other) and provides no way for users to clear the legacy field. This makes import impossible for any affected project; no Terraform config can satisfy both the provider schema and the live state.

**Affected resource:** `vercel_project` (import path).

### To Reproduce

1. Confirm a victim project exists — GET returns both fields populated:

   ```bash
   curl -s -H "Authorization: Bearer $VERCEL_TOKEN" \
     "https://api.vercel.com/v9/projects/$PROJECT_ID" | \
     jq '{serverlessFunctionRegion, resourceConfig}'
   ```

   ```json
   {
     "serverlessFunctionRegion": "sfo1",
     "resourceConfig": {
       "functionDefaultRegions": ["sfo1"],
       "functionDefaultMemoryType": "standard",
       "elasticConcurrencyEnabled": true
     }
   }
   ```

2. Declare + import in Terraform:

   ```hcl
   resource "vercel_project" "legacy" {
     name      = "my-project"
     framework = "nextjs"
     # No region fields declared
   }

   import {
     id = "prj_xxx"
     to = vercel_project.legacy
   }
   ```

3. Run `terraform plan` — import fails with the two errors quoted above.

### Terraform and Vercel Provider Version

```console
Reproduced via Pulumi's bridge of this provider (`@pulumiverse/vercel@4.6.1`) rather than Terraform directly. Happy to reproduce via a Terraform-native setup if useful — the repro above is provider-schema-level so the same ConflictsWith rule fires regardless of front-end.

Vercel API version: tested against both `/v9/projects/` and `/v10/projects/` — identical behavior.
```

### Additional Details

**Why client-side workarounds fail:** I tested four REST-API variants to try clearing the legacy field before import. None work:

| Variant | Result |
|---|---|
| `PATCH {"serverlessFunctionRegion": null}` | 200 OK, field unchanged — API silently ignores the null |
| `PATCH {"serverlessFunctionRegion": ""}` | 400 — "If defined, the Serverless Function Region must be a string of 4 characters" |
| Both fields in one body (`null` + `["sfo1"]`) | 400 — "Cannot specify both" |
| Write modern field alone (`resourceConfig.functionDefaultRegions: ["iad1"]`) | 200, but BOTH fields flip to `iad1` — Vercel auto-mirrors |

Same behaviors against `/v9/` and `/v10/`. Dashboard re-save of the Functions region doesn't clear the legacy field either. Vercel's docs team has confirmed on the record (support ticket in parallel) that there's no public path — API or UI — to clear the legacy field on pre-migration projects.

**Root cause (believed):** The `ConflictsWith` constraint models the API's *write* semantics (where you can set one field or the other, not both) rather than its *read* semantics (where both coexist on pre-migration projects). The provider therefore cannot ingest live state that the API itself produces.

**Proposed fix:** Drop the `ConflictsWith` constraint between these two fields. Mark `serverless_function_region` as `Computed` so Vercel's auto-mirroring doesn't drift Terraform state. On Update, prefer writing `resource_config.function_default_regions` alone (matches the provider's forward-looking intent and avoids the API's own "cannot specify both" rule on writes).

**Environment:**
- Reproduced via Pulumi's bridge (`@pulumiverse/vercel@4.6.1`). Happy to reproduce with Terraform directly if needed — the failure is schema-level, so the frontend shouldn't matter.
- Vercel API: `/v9/projects/` and `/v10/projects/` both tested.

**Context for triage:** We have two projects stuck in this state in our team. Skipping import leaves them outside IaC, which defeats the purpose of adopting the provider. Happy to supply API payloads, stored-state samples, or validate a fix-branch PR.

haseebrj17:
Fixes #475.

## Root cause

The Vercel API returns `serverlessFunctionRegion` as a **derived, backwards-compatibility alias** for `resourceConfig.functionDefaultRegions[0]` — confirmed by Vercel Support on the support ticket referenced in #475:

> `serverlessFunctionRegion` is not behaving like an independently stored field that can be cleared directly. Instead, it is returned by the API as a backwards-compatibility alias derived from `resourceConfig.functionDefaultRegions[0]`. […] we do not have a programmatic way to permanently clear `serverlessFunctionRegion` while `resourceConfig.functionDefaultRegions` is set.

The provider's reciprocal `ConflictsWith` validators modeled the API's *write* semantics (where you can set one field or the other, not both) but not its *read* semantics (where both coexist on any project created or touched via the modern path). The result: projects that the API itself produces fail the provider's own import-time validation.

## Why this reproduced through Pulumi but not plain Terraform

`ConflictsWith` from `terraform-plugin-framework-validators` checks `req.ConfigValue` — config, not state. In pure Terraform, `import { … }` brings state in without populating config, so the config slot for both attributes is null and the validator does not fire. Pulumi's Terraform bridge, however, reifies imported state back into Pulumi inputs, and those inputs become the effective config for the next validation pass. Both fields land in config with non-null values, and both reciprocal errors fire — which is exactly what @haseebrj17 hit.

## Change

- **Drop both `ConflictsWith` validators** on `serverless_function_region` and `resource_config.function_default_regions`. The write path in `toClientResourceConfig` already collapses the two into a single `functionDefaultRegions` array and prefers the modern value, so removing the config-level conflict is safe on `Create`/`Update`.
- **Add a `ModifyPlan` warning** (not an error) when a user authors both fields in config. The warning reads from `req.Config`, not `req.Plan`, so it only fires for user-authored dual-declaration and *not* for state round-trips during import.
- **Remove the unused `ServerlessFunctionRegion` field** from `CreateProjectRequest` and `UpdateProjectRequest` in `client/project.go`. The provider never populated these — `toClientResourceConfig` always folded the legacy value into `functionDefaultRegions`. This is strictly cleanup to keep future contributors from assuming the legacy field is writable over the v9 PATCH endpoint (the API rejects it anyway).

## Tests

- New: `TestAcc_ProjectImportDualRegion` — creates a project via `resource_config.function_default_regions = [\"sfo1\"]`, imports the resulting dual-populated state, and asserts an empty follow-up plan. This is the regression guard for #475: the Pulumi-bridge failure reduced to plain Terraform.
- Updated: step 1 of `TestAcc_ProjectFunctionDefaultRegions`, which previously asserted `ConflictsWith` fired. It now asserts the dual-field config succeeds and the modern field wins in state.

## Behavior change to call out for reviewers

Previously, declaring both fields in config was a hard error at plan time. It is now **permitted**, with the modern field winning silently plus a plan-time deprecation warning. This matches Vercel's own API semantics (where the legacy field is a derived alias and setting both in one PATCH is rejected by the API itself).

## No-ops

- No client-side clearing of `serverlessFunctionRegion` is possible — Vercel Support confirmed the field cannot be cleared server-side while `functionDefaultRegions` is set. This PR does not attempt any workaround; it simply stops rejecting the state the API returns.
- `data_source_project.go` already declares `serverless_function_region` as `Computed` with no `ConflictsWith`, so data sources were never affected.
- `docs/resources/project.md` is unchanged — `tfplugindocs` does not render `ConflictsWith` in markdown, only Description + `Deprecated` marker, and both are unchanged.

## Test plan

- [ ] Run `task test -- -run 'TestAcc_ProjectImportDualRegion'` against a Vercel test team — expect green.
- [ ] Run `task test -- -run 'TestAcc_ProjectFunctionDefaultRegions'` — expect green (updated step 1).
- [ ] Spot-check a fresh `terraform import` against a pre-migration project with both fields populated — expect clean state + empty follow-up plan.
- [ ] Manually verify the deprecation warning renders on `terraform plan` when both fields are in config.

### DIFF
--- a/client/project.go
+++ b/client/project.go
@@ -57,7 +57,6 @@ type CreateProjectRequest struct {
 	PreviewDeploymentSuffix           *string               `json:"previewDeploymentSuffix"`
 	PublicSource                      *bool                 `json:"publicSource"`
 	RootDirectory                     *string               `json:"rootDirectory"`
-	ServerlessFunctionRegion          string                `json:"serverlessFunctionRegion,omitempty"`
 	ResourceConfig                    *ResourceConfig       `json:"resourceConfig,omitempty"`
 	EnablePreviewFeedback             *bool                 `json:"enablePreviewFeedback,omitempty"`
 	EnableProductionFeedback          *bool                 `json:"enableProductionFeedback,omitempty"`
@@ -327,7 +326,6 @@ type UpdateProjectRequest struct {
 	PreviewDeploymentSuffix              *string                         `json:"previewDeploymentSuffix"`
 	PublicSource                         *bool                           `json:"publicSource"`
 	RootDirectory                        *string                         `json:"rootDirectory"`
-	ServerlessFunctionRegion             string                          `json:"serverlessFunctionRegion,omitempty"`
 	VercelAuthentication                 *VercelAuthentication           `json:"ssoProtection"`
 	PasswordProtection                   *PasswordProtectionWithPassword `json:"passwordProtection"`
 	TrustedIps                           *TrustedIps                     `json:"trustedIps"`
--- a/vercel/resource_project.go
+++ b/vercel/resource_project.go
@@ -120,10 +120,6 @@ At this time you cannot use a Vercel Project resource with in-line ` + "`environ
 				Description:        "The region on Vercel's network to which your Serverless Functions are deployed. It should be close to any data source your Serverless Function might depend on. A new Deployment is required for your changes to take effect. Please see [Vercel's documentation](https://vercel.com/docs/concepts/edge-network/regions) for a full list of regions.",
 				Validators: []validator.String{
 					validateServerlessFunctionRegion(),
-					stringvalidator.ConflictsWith(
-						path.MatchRoot("serverless_function_region"),
-						path.MatchRoot("resource_config").AtName("function_default_regions"),
-					),
 				},
 			},
 			"node_version": schema.StringAttribute{
@@ -582,10 +578,6 @@ At this time you cannot use a Vercel Project resource with in-line ` + "`environ
 						Validators: []validator.Set{
 							setvalidator.ValueStringsAre(
 								validateServerlessFunctionRegion(),
-							),
-							setvalidator.ConflictsWith(
-								path.MatchRoot("serverless_function_region"),
-								path.MatchRoot("resource_config").AtName("function_default_regions"),
 							),
 						},
 					},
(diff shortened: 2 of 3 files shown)


dglsparsons:
This appears to be an AI drive-by, and removing ConflictsWith is no bueno.

PR #486, opened 2026-04-28
fix: skip empty build_machine_type so adopted projects can update

mehulkar:
## Summary

When a project's API response had no `buildMachineType` value, the provider's Read path stored it in state as `types.StringValue("")` — a known empty string rather than null. On subsequent updates, that planned value was forwarded to the API as `buildMachineType: ""`, which the API rejects (allowed values are `null`, `enhanced`, `turbo`, `standard`, `elastic`):

```
unexpected error: bad_request - Invalid request:
`resourceConfig.buildMachineType` should be equal to one of the
allowed values ", enhanced, turbo, standard, elastic".
```

This affected projects adopted with no build machine type set, even when `build_machine_type` was in `lifecycle.ignore_changes` — `ignore_changes` substitutes the state value into the plan, and the state value was `""` rather than null.

`toClientResourceConfig` now skips `BuildMachineType`/`BuildMachineSelection` when the value is null or empty, so the empty-string case maps to the API's "omit the field" behavior.

## Test plan

- [x] New unit test `TestToClientResourceConfig_BuildMachineType` covers null, empty, unknown, `enhanced`, and `elastic` inputs and fails on the unfixed code in the empty-string case
- [ ] Re-run the failing terraform plan against an adopted project to confirm it now succeeds

### DIFF
--- a/vercel/resource_project.go
+++ b/vercel/resource_project.go
@@ -1439,7 +1439,11 @@ func (r *ResourceConfig) toClientResourceConfig(ctx context.Context, onDemandCon
 		}
 		resourceConfig.ElasticConcurrencyEnabled = onDemandConcurrentBuilds.ValueBoolPointer()
 	}
-	if !buildMachineType.IsUnknown() {
+	// The API rejects an explicit `buildMachineType: ""` (allowed values are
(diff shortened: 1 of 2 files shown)


mehulkar:
Superseded by #487, which combines this fix with the elastic request-payload fix.

PR #494, opened 2026-04-29
fix shared env update false boolean

dglsparsons:


dglsparsons:
Closing this because the pointer/omitempty change can send computed state instead of only user-explicit config. We'll handle any explicit false behavior in a separately designed change.

PR #525, opened 2026-06-16
Remove publicSource from project requests

swarnava:
The Vercel API no longer accepts publicSource on project create or update requests. (https://github.com/vercel/vercel/pull/16675)  Stop serializing the removed field so projects can be created even when public_source is not configured.

Deprecate public_source in the resource and data source schemas, and preserve any configured deprecated value in state when the API omits it to avoid Terraform consistency churn.


swarnava:
Fixing in https://github.com/vercel/terraform-provider-vercel/pull/527

PR #557, opened 2026-07-15
Add Vercel KMS support

marcgreenstock:
## Summary

Adds Terraform support for managing Vercel KMS, which previously had no
provider coverage — issuers, their signing keys, and project grants could only
be managed via the dashboard or API. This lets users manage the full KMS
lifecycle as code.

### `vercel_kms_issuer` (resource + data source)

JWT issuer with a JWKS endpoint; `name` is mutable, key material is
immutable/computed. Supports importing an external private key.

```hcl
resource "vercel_kms_issuer" "example" {
  name      = "my-issuer"
  algorithm = "RS512"
}

# Import an externally-generated key instead of having Vercel generate one.
resource "vercel_kms_issuer" "external" {
  name       = "my-external-issuer"
  import_key = file("${path.module}/private-key.pem")
  key_id     = "my-key-id"
}

data "vercel_kms_issuer" "example" {
  id = "iss_xxxxxxxxxxxxxxxxxxxxxxxxxxxx"
}
```

### `vercel_kms_project_grant`

Authorizes a Vercel project to request signed tokens in the listed
environments, optionally with additional token claims.

```hcl
resource "vercel_kms_project_grant" "example" {
  issuer_id    = vercel_kms_issuer.example.id
  project_id   = vercel_project.example.id
  environments = ["production", "preview"]

  token_claims = jsonencode({
    aud = "https://example.com"
  })
}
```

### `vercel_kms_signing_key`

Models an imperative key rotation via a `keepers` trigger; keys retire on the
next rotation (no single-key delete API).

```hcl
resource "vercel_kms_signing_key" "example" {
  issuer_id = vercel_kms_issuer.example.id

  # Change any value in keepers to rotate in a new signing key.
  keepers = {
    rotated_at = "2024-01-01"
  }
}
```

## Validation

- KMS acceptance tests run under the standard `TF_ACC` harness like other
  resources. The issuer test covers create → import → rename; the signing-key
  test asserts that changing `keepers` produces a new key id.

Made with [Cursor](https://cursor.com)


client/kms.go
```diff
@@ -0,0 +1,313 @@
...
+		url = fmt.Sprintf("%s?teamId=%s", url, c.TeamID(teamID))
+	}
+	return url
+}
+
+type CreateKMSIssuerRequest struct {
+	Name      string `json:"name"`
+	Algorithm string `json:"algorithm,omitempty"`
+	ImportKey string `json:"importKey,omitempty"`
+	KeyID     string `json:"keyId,omitempty"`
```
dglsparsons: [Doug's Clanker] When key_id is configured, this sends keyId, but both KMS endpoints accept importKeyId and reject additional properties, so issuer creation and key rotation fail with a 400. Use the importKeyId JSON tag in both request structs.

client/kms.go
```diff
@@ -0,0 +1,313 @@
...
+	TeamID    string `json:"-"`
+}
+
+func (c *Client) CreateKMSIssuer(ctx context.Context, request CreateKMSIssuerRequest) (i KMSIssuer, err error) {
+	url := c.kmsIssuersURL(request.TeamID)
+	body := string(mustMarshal(request))
+	tflog.Info(ctx, "creating kms issuer", map[string]any{
+		"url":  url,
+		"body": body,
```
dglsparsons: [Doug's Clanker] When import_key is set, this writes the full private key into Terraform provider logs. Remove the request body from the CreateKMSIssuer and RotateKMSIssuerKey log fields, or redact importKey before logging.

client/kms.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+	return p, nil
+}
+
+type UpdateKMSProjectGrantRequest struct {
+	IssuerID     string          `json:"-"`
+	TeamID       string          `json:"-"`
+	ProjectID    string          `json:"-"`
+	Environments []string        `json:"environments,omitempty"`
+	TokenClaims  json.RawMessage `json:"tokenClaims,omitempty"`
```
dglsparsons, line 274: [Doug's Clanker] Removing token_claims omits tokenClaims, so the API treats the update as a no-op and keeps the saved claims. The API requires tokenClaims: null to clear them; preserve that distinction in this request.

vercel/resource_kms_project_grant.go
```diff
@@ -0,0 +1,349 @@
...
+			},
+			"environments": schema.SetAttribute{
+				Required:    true,
+				ElementType: types.StringType,
+				Description: "The environments in which the project may request signed tokens (for example `production`, `preview`).",
+				Validators: []validator.Set{
+					setvalidator.SizeAtLeast(1),
+				},
+			},
+			"token_claims": schema.StringAttribute{
```
dglsparsons, line 109: [Doug's Clanker] Invalid token_claims reaches json.RawMessage unchecked, and mustMarshal then panics instead of returning a Terraform diagnostic. Validate that this is a JSON object at plan time, including rejecting the empty string.

vercel/kms_shared.go
```diff
@@ -0,0 +1,160 @@
...
+		"certificate_pem":        types.StringType,
+		"created_at":             types.StringType,
+		"updated_at":             types.StringType,
+		"revoke_at":              types.StringType,
+	}
+}
+
+func kmsSigningKeyValue(key client.KMSSigningKey) kmsSigningKeyModel {
+	return kmsSigningKeyModel{
+		KeyID:                types.StringValue(key.KeyID),
```
dglsparsons, line 50: [Doug's Clanker] For an imported key, key.KeyID is the server-minted addressable ID, while importKeyId is the JWT/JWKS kid. This makes signing_keys[*].key_id return a different value from the one its schema promises, so callers can select the wrong kid. Describe this as the addressable key ID or expose the JWT kid separately.

vercel/resource_kms_signing_key.go
```diff
@@ -0,0 +1,341 @@
...
+				PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplace()},
+			},
+			"algorithm": schema.StringAttribute{
+				Computed:      true,
+				Description:   "The signing algorithm of the key.",
+				PlanModifiers: []planmodifier.String{stringplanmodifier.UseStateForUnknown()},
+			},
+			"status": schema.StringAttribute{
+				Computed:    true,
+				Description: "The status of the key, either `active` or `revoking`.",
```
dglsparsons: [Doug's Clanker] A newly rotated key is returned as pending until its public key propagates, so this documented set omits a value callers receive during normal operation. Include pending in the status descriptions for the signing-key resource and issuer signing_keys.

vercel/kms_shared.go
marcgreenstock: I moved it all to `importKeyId`. `keyID` in this case is always the addressable ID.

dglsparsons (approved):

PR #564, opened 2026-07-24
Add destructive operation warning for resource

rmainwork:
Since this resource has the potential to be a destructive operation - I added a note to the docs to reflect that and hopefully prevent someone from shooting their hand of (as I nearly did :grin:)


dglsparsons:
Superseded by #565, which moves the warning into the resource schema and clarifies the behavior to match the current implementation. Thanks for raising the destructive-operation concern.

PR #577, opened 2026-08-07
Add vercel_oauth_app_permissions resource

TooTallNate:
## What

Adds a dedicated **`vercel_oauth_app_permissions`** resource granting Vercel REST API permissions (e.g. `read:team`, `read:project`, `read-write:project`, `read:deployment`, `read-write:deployment`) to a Sign in with Vercel app's tokens — the piece that lets an app act on the platform on the signed-in user's behalf, consented at sign-in. This is the same field the dashboard's Permissions page sets via `PATCH /v1/oauth-apps/:clientId`, so an app can be provisioned end-to-end by Terraform with zero dashboard involvement.

> **Design history**: the first revision put `permissions` on `vercel_oauth_app` itself, which required a create-then-PATCH-then-delete-on-failure dance (the create endpoint doesn't accept permissions). Review rightly called that hack dangerous; per follow-up discussion, the grant is now its own resource mapping 1:1 onto the PATCH endpoint, and `vercel_oauth_app`'s Create is a single atomic API call again.

### Semantics

- One grant resource per app; it owns the app's **whole** permission set
- **Create/Update** PATCH only `{ permissions }`; **Delete revokes** (PATCHes `[]`) rather than merely forgetting; **Read** removes itself when the app is gone or has no grants; importable as `[team_id/]client_id`
- **Clobber-proof by construction**: the general `UpdateOAuthAppRequest` deliberately has *no* permissions field (the API preserves permissions when the field is absent), and `UpdateOAuthAppPermissions` sends *only* that field — the two resources can't stomp on each other
- No client-side value enum: the server's evolving permission list is the source of truth
- Verified against production that setting permissions works with a plain team-owner token

### Testing

Acceptance tests pass against the production API:

```
--- PASS: TestAcc_OAuthAppPermissionsResource (5.03s)  # grant + API read-back, update, import,
                                                       # then remove the grant while KEEPING the app
                                                       # → proves destroy revokes
--- PASS: TestAcc_OAuthAppResource (3.12s)
--- PASS: TestAcc_OAuthAppClientSecretResource (2.41s)
```

Unit tests, `staticcheck`, `tfproviderlint -R018=false`, `gofmt -s`, `go vet` clean; docs regenerated.

### Follow-up

The deeper fix — the create endpoint accepting `permissions` directly — remains an `api-login-oauth2` change worth making, but nothing here depends on it anymore.

## Motivation

Follow-up to #576: vercel/infra#32272 provisions durabench's Sign in with Vercel app via Terraform, and the API-permission grant was the one remaining dashboard step. With this it's fully declarative:

```hcl
resource "vercel_oauth_app_permissions" "durabench_sign_in" {
  oauth_app_id = vercel_oauth_app.durabench_sign_in.id
  permissions  = ["read:team", "read:project", "read-write:project", "read:deployment", "read-write:deployment"]
}
```


vercel/resource_oauth_app.go
```diff
@@ -241,6 +264,34 @@ func (r *oauthAppResource) Create(ctx context.Context, req resource.CreateReques
...
+		})
+		if err != nil {
+			_ = r.client.DeleteOAuthApp(ctx, out.ClientID, out.TeamID)
+			resp.Diagnostics.AddError(
+				"Error creating OAuth App",
+				"Could not grant permissions to the created OAuth App (the app has been deleted), unexpected error: "+err.Error(),
+			)
+			return
+		}
+	}
```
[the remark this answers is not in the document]
TooTallNate: Both points were right — fixed in d7f12c1.

**null arrays**: rather than patching the one call site, `UpdateOAuthApp` now normalizes nil `RedirectURIs`/`Scopes`/`Permissions` to empty slices at the client boundary — the request struct intentionally lacks `omitempty` (explicit nulls clear the nullable URL fields), so *any* caller passing a nil slice would have serialized `null` where the API expects an array. The client-secret acceptance test now creates its app with `permissions` and **no** `redirect_uris`, pinning exactly the flagged combination against the production API (passes).

**rollback honesty**: the delete error is no longer discarded — the diagnostic now reports whether the rollback actually succeeded, and when it didn't, names the client id so the leaked app can be deleted manually instead of falsely claiming it's gone.

vercel/resource_oauth_app.go
```diff
@@ -241,6 +264,40 @@ func (r *oauthAppResource) Create(ctx context.Context, req resource.CreateReques
 		return
 	}
 
+	// The create endpoint does not accept permissions — grant them with an
+	// immediate follow-up update. On failure the half-configured app is deleted
+	// so a failed create never leaves an unmanaged app behind.
```
jarneson: this really needs to change in the api surface, these sorts of hacks are VERY dangerous

jarneson (approved):

vercel/resource_oauth_app.go
TooTallNate: Agreed — reworked in c28e717 per your Slack suggestion: permissions are now a dedicated `vercel_oauth_app_permissions` resource that maps 1:1 onto the PATCH endpoint, and the create-then-PATCH-with-rollback hack is gone entirely — `vercel_oauth_app`'s Create is a single atomic API call again.

Two properties worth calling out: the general update request deliberately has **no** permissions field (the API preserves permissions when absent) and the permissions call sends **only** that field, so the two resources can't clobber each other by construction. And destroying the grant resource *revokes* (PATCHes `[]`) rather than merely forgetting — there's an acceptance step that removes the grant while keeping the app and asserts the API reports zero permissions.

Fully agreed the real fix is the create endpoint accepting `permissions` on the API side - happy to send that separately; nothing here depends on it anymore.

client/oauth_app.go
```diff
@@ -35,6 +35,7 @@ type OAuthApp struct {
 	HomePageURI       string                         `json:"homePageUri"`
 	RedirectURIs      []string                       `json:"redirectUris"`
 	Scopes            []string                       `json:"scopes"`
+	Permissions       []string                       `json:"permissions"`
```
dglsparsons, line 38: If the API models permissions as part of the oauth_app, why do we have this as a separate resource? 👀

dglsparsons (changes requested):
just to block this while FF state, etc. is figured out.

PR #580, opened 2026-08-10
Fix typo in Edge Config item import docs

dglsparsons:
Fixes a typo in the Edge Config Item import example and its generated documentation.

Supersedes #578 with the same correction recreated as a cryptographically signed, verified commit.


ecklf (approved):
Stamp

PR #582, opened 2026-08-11
Add vercel_alert_rule resource and data source

jsvana:
## Summary

Adds support for managing [Vercel alert rules](https://vercel.com/docs/alerts) in Terraform, so teams can configure monitoring as code instead of clicking through the Observability dashboard or scripting `vercel alerts rules`.

Alert rules decide which anomalies and custom Observability thresholds raise an alert for a team or a set of projects, and who is subscribed to the alerts raised.

* New resource `vercel_alert_rule` (full CRUD + import)
* New data source `vercel_alert_rule`
* New client layer against `/alerts/v2/alert-rules`

Both kinds of rule the API supports are covered:

```hcl
# Built-in anomaly rule, scoped with an OData project filter
resource "vercel_alert_rule" "checkout_errors" {
  name           = "Checkout 5xx anomalies"
  project_filter = "projectId in ('${vercel_project.example.id}')"
  alert_types = [{
    type   = "error_anomaly"
    filter = "statusGroup eq '5xx' and route eq '/api/checkout'"
  }]
  sensitivity_level = 4
}

# Custom alert rule: alert when the hourly 5xx rate goes above 5%
resource "vercel_alert_rule" "checkout_error_rate" {
  name        = "Checkout error rate"
  project_id  = vercel_project.example.id
  alert_types = [{ type = "custom_alert" }]

  custom_alert = {
    event = "incomingRequest"
    rollups = {
      errors   = { measure = "count", aggregation = "sum", filter = "httpStatus ge 500" }
      requests = { measure = "count", aggregation = "sum" }
    }
    granularity       = "1h"
    trigger_type      = "threshold"
    trigger_operator  = "gt"
    trigger_threshold = 0.05
    min_threshold     = 20
    formula           = { operator = "divide", left = "errors", right = "requests" }
  }
}
```

## Design notes

A few API details drove the shape of the schema.

**The Observability query is modelled as attributes, not a JSON string.** The API carries it as an escaped JSON string in `customAlert.queryJsonString`. Rather than making users hand-assemble that with `jsonencode`, `custom_alert` exposes `event`, `rollups`, `group_by`, `filter`, and `granularity` as ordinary attributes and the client serializes them. This also avoids the perpetual diffs a round-tripped JSON string invites. `granularity` is expressed as `5m`/`1h`/`1d` rather than the API's `{"hours": 1}` object.

**The query `scope` is derived, not configured.** The API rejects a custom alert query with no `scope`, and the Vercel CLI injects one client-side. The client does the same, from the rule's team and project, so it is not duplicated in config.

**`project_filter` vs `project_id`.** The API has a single `projectId` field that holds an OData filter (`projectId in ('prj_123')`) for built-in rules but a raw project ID for custom alert rules. Collapsing those into one Terraform attribute would be confusing, so they are separate attributes mapped back on read based on the kind of rule. `ValidateConfig` prevents mixing them up, along with the other type-dependent combinations (`custom_alert` requires the `custom_alert` alert type and a `project_id`; `min_threshold` only applies to `anomaly` triggers or triggers with a `formula`; `formula` sides must name real `rollups` keys).

**Updates send explicit nulls.** The API clears an optional field when it is sent as `null`, so the update payload always includes `projectId`, `sensitivityLevel`, and both autosubscribe flags. `customAlert` is the exception — the API rejects `"customAlert": null` — so it is omitted when unset, and adding or removing `custom_alert` replaces the rule rather than updating it.

**Sweep.** Alert rules are not tied to a project, so `deleteAllProjects` does not clean them up. `sweep` now deletes leftover rules, skipping the team's Vercel-managed default rule (`isDefault`), which cannot be deleted.

Slack/webhook subscription for a rule is done outside Terraform (`/vercel subscribe <team-id> alerts +rule:<rule-id>`), so the `notifications` field is deliberately not exposed; the rule `id` is available for that. `sensitivity_level` is validated to the API's 1–5 range.

## Testing

`task lint` is clean (staticcheck, tfproviderlint, gofmt, go vet), and the unit tests pass.

* `client/alert_rule_test.go` — request/response shapes over `httptest`, including that create omits unset optionals, that update sends explicit nulls but omits `customAlert`, that the query scope is derived (and an explicit one is preserved), and 404 handling.
* `vercel/resource_alert_rule_unit_test.go` — schema expectations, granularity round-trip, `projectId` mapping onto `project_filter` vs `project_id`, and a model round-trip through the client types.
* `vercel/resource_alert_rule_test.go` — acceptance tests for team-wide and project-scoped built-in rules (create, import, then an update that renames the rule and clears `sensitivity_level`, `project_filter`, and an alert type filter), plus the data source, and a separate test for custom alerts.

Beyond that, I ran the resource end-to-end against a real team with a locally built provider to confirm `terraform plan` is empty after apply, after update, and after import — create, in-place update with fields cleared, import round-trip, data source read, and destroy all behaved, and each of the validation errors above fires. All test resources were removed afterwards.

One coverage gap worth flagging: `TestAcc_AlertRuleResourceCustomAlert` is skipped unless `VERCEL_TERRAFORM_TESTING_OBSERVABILITY_PLUS` is set, because the shared testing team has no Observability Plus subscription and the API rejects custom alert rules without one (`forbidden - Access to custom alerts requires an Observability Plus subscription`). Set that variable against a team with a non-zero custom alert allowance to run it. The custom alert request shape was still validated by the API up to its quota check, and the conversions are unit-tested in both directions, but that acceptance test has not been run green against a real team.


dglsparsons:
See slack discussion.

PR #583, opened 2026-08-14
Add import support for Vercel webhooks

Relequestual:
I'm in the process of importing our Vercel setup into Terraform as much as possible, and I got stuck. I'd just made a bunch of replacement webhooks, and I didn't want to do it again. So I patched the provider locally! But there's no reason everyone else shouldn't get it too.

`vercel_webhook` cannot import an existing webhook, which prevents Terraform from adopting an existing signed webhook without recreating it.

This adds import support for `team_id/webhook_id`, or `webhook_id` when the provider has a default team. Import reads the existing endpoint, events, project IDs, team, and webhook ID.

Vercel does not return an existing webhook signing secret. Imported state therefore records `secret` as null; this does not change how Terraform-created webhooks behave.

Includes unit coverage for valid and malformed import identifiers, missing webhooks, and API failures, plus acceptance import coverage, generated documentation, and an import example.

I might be a little slow to see any feedback. I'm 100% fine if any changes are required that you're happy to make if it means after it will be merged.


dglsparsons:
Merged in #584 rather than this PR. Released in v5.11.1 of the provider. 

Thanks for raising this!

PR #585, opened 2026-08-18
Keep project IDs in firewall config state

dglsparsons:
`vercel_firewall_config` drops the project ID after API reads and writes because the API response does not include it. This leaves the Terraform resource ID as `team_id/` instead of `team_id/project_id`.

Preserve the project ID from the request on both GET and PUT responses so Terraform state always records the complete resource ID.


ecklf (approved):
Stamp

PR #586, opened 2026-08-18
Update edited firewall rules in place

dglsparsons:
When a firewall rule's name and body change together, its computed ID is unknown in the proposed plan. The provider can mistake the edit for a removal plus insertion, delete the existing rule, then fail to insert the replacement if API validation rejects it.

Carry IDs from state into the proposed plan by matching stable rule details first and list order only as a final fallback. Existing rules are then updated in place, so a rejected edit leaves the active rule unchanged; genuinely new rules still receive IDs after creation.


ecklf (approved):
Stamp

PR #588, opened 2026-08-21
Add project tracing resource

lukahartwig:
Adds the ability to [configure project tracing settings](https://vercel.com/changelog/always-on-tracing-for-production-and-preview-traffic) via Terraform provider.

Example

```
  name = "example"
}

resource "vercel_project_tracing" "example" {

  sampling_rules = [{
    rate         = 0.2
    environment  = "production"
    request_path = "/api"
  }]
}
```

Deleting the resource disables tracing for the project. An empty or omitted `sampling_rules` array means no filter is applied and 100% traces are retained.


vercel/resource_project_tracing.go
```diff
@@ -0,0 +1,269 @@
...
+type ProjectTracing struct {
+	ID            types.String `tfsdk:"id"`
+	ProjectID     types.String `tfsdk:"project_id"`
+	TeamID        types.String `tfsdk:"team_id"`
+	Enabled       types.Bool   `tfsdk:"enabled"`
+	SamplingRules types.List   `tfsdk:"sampling_rules"`
+}
+
+func responseToProjectTracing(ctx context.Context, out client.ProjectTracing, preferredSamplingRules types.List) (ProjectTracing, diag.Diagnostics) {
+	samplingRules, diags := traceDrainSamplingRulesFromAPI(ctx, out.SamplingRules, preferredSamplingRules)
```
lucleray, line 121: i think this creates perpetual drift when tracing is disabled. the endpoint returns `sampling: []` for `enabled = false`, so a config that keeps non-empty `sampling_rules` gets written back as `[]`; the next plan proposes the same update again. could we preserve `preferredSamplingRules` while disabled (or reject rules when disabled) and cover the enabled → disabled transition?

lucleray (commented):
one issue inline

vercel/resource_project_tracing.go
```diff
@@ -0,0 +1,269 @@
...
+			"team_id": schema.StringAttribute{
+				Optional:      true,
+				Computed:      true,
+				Description:   "The ID of the team the Project exists under. Required when configuring a team resource if a default team has not been set in the provider.",
+				PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplaceIfConfigured(), stringplanmodifier.UseNonNullStateForUnknown()},
+			},
+			"enabled": schema.BoolAttribute{
+				Required:    true,
+				Description: "Whether tracing is enabled for the Project.",
+			},
```
dglsparsons: Should this resources existence not imply tracing is enabled?

lukahartwig: Hmm, yeah. The idea was that the config exists but tracing is disabled. I was copying the pattern from cron.

That being said, we don't have a UI state for this at the moment, so I can remove this and we can introduce this later.

vercel/resource_project_tracing.go
lukahartwig: I removed `enabled` from the config for now so it's no longer possible to declare a `enabled = false` and `sampling: [{ ... }]` resource.

vercel/resource_project_tracing.go
lukahartwig: Done

lucleray (approved):
looks good, thanks! 👍

dglsparsons (approved):

PR #589, opened 2026-08-21
Add AI Gateway API key resource

TooTallNate:
## Summary

- add a `vercel_ai_gateway_api_key` resource for provisioning AI Gateway API keys through `/v1/api-keys`
- preserve the one-time `api_key_string` as sensitive Terraform state while supporting refresh, import, and deletion by stable key ID
- support `expires_at` and `project_id` (create-only, force replacement per the API contract)
- support `ai_gateway_quota` (spend budget): upserted in-place via `PATCH /v1/api-keys/{id}/quota`, archived when removed from configuration, and read back through the list endpoint's quota decoration for drift detection
- renaming a key is an in-place update (`PATCH /v1/api-keys/{id}`), so the secret is preserved across renames
- add client unit tests, acceptance lifecycle tests, examples, and generated documentation

## Notes

- quota management is gated by the `can-create-ai-gateway-api-key-quotas` team feature flag; the quota acceptance test requires it enabled for the testing team
- `include_byok_in_quota` is intentionally omitted (feature not yet launched, per @shaper)
- key metadata (`environment`, `spendAttribution`, `zdr`, `bypassAll`) is left as a follow-up

## Validation

- validated request, response, team scoping, and lifecycle behavior against the implementations in `vercel/api` and `vercel/ai-gateway`
- `go test ./client ./vercel -run 'AIGatewayAPIKey|TestAllResourcesHaveIDAttribute'`
- `go vet ./...`
- `staticcheck -tags it ./...`
- `tfproviderlint -R018=false ./...`
- generated docs with `tfplugindocs v0.25.0`

The live acceptance tests require `VERCEL_API_TOKEN` and `VERCEL_TERRAFORM_TESTING_TEAM` and are left to the configured CI environment.


shaper (commented):
This is explicitly creating an AI Gateway API Key, but there are also other API keys -- sandbox, botid I believe, maybe others.

IIRC the `purpose` is what makes it gateway specific, and there are restrictions on what services can create with that purpose.

We should check with accounts team and get their steering. We could update this PR and terraform action names, etc. to make it explicitly AI Gateway API keys to keep it focused and clear -- would still check with them to validate the thinking.

@markandrus to guide us further

shaper (approved):
**Note: everything below is non-blocking IMO — all of it can land as follow-ups.** One open question before/at merge: does this ship publicly (Terraform Registry + `docs/resources/api_key.md` going live) as soon as the next provider release is cut? I'm not sure of the release cadence here — if it does go public immediately, do we want a changelog entry / announcement, or should we hold the release until quota support lands?

Nice minimal v1 — create/read/delete/import lifecycle, one-time `api_key_string` handling, and tests all look solid. But as written, the resource can only set `name` and `purpose`, which is a small fraction of what `POST /v1/api-keys` accepts. It can't express any of the quota/expiration options that `vercel ai-gateway api-keys create` exposes today.

**How a CLI create flows through the stack:**

```
vercel ai-gateway api-keys create
  --name --budget --refresh-period --include-byok --alert-thresholds --expiration
        |
        v
  CLI builds body: { purpose, name?, expiresAt?, aiGatewayQuota? }
  (cli: src/commands/ai-gateway/api-keys-create.ts,
        src/util/ai-gateway/quota.ts, src/util/ai-gateway/api-keys.ts)
        |
        v
  POST /v1/api-keys  (api: services/api-keys/src/endpoints/create-api-key.ts)
        |
        +--> creates key record        (api: services/api-keys/src/lib/create-api-key.ts)
        |
        +--> if aiGatewayQuota set:
             HMAC key secret, then
             POST quota to AI Gateway  (api: services/api-keys/src/lib/ai-gateway/create-quota.ts)
                    |
                    v
             AI Gateway quotas endpoint (ai-gateway: app/v1/quotas/route.ts)
```

**Full request body accepted by the endpoint** (see `BodySchema` in `api/services/api-keys/src/endpoints/create-api-key.ts`):

| Body field | CLI flag | This PR |
|---|---|---|
| `purpose` (required) | implicit `ai-gateway` | ✅ |
| `name` | `--name` | ✅ |
| `expiresAt` (ms timestamp) | `--expiration 7d/30d/60d/90d/1y/none` | ❌ |
| `aiGatewayQuota.limitAmount` (required within quota, min 1) | `--budget` | ❌ |
| `aiGatewayQuota.refreshPeriod` (`daily/weekly/monthly/none`) | `--refresh-period` | ❌ |
| `aiGatewayQuota.includeByokInQuota` | `--include-byok` | ❌ |
| `aiGatewayQuota.alertThresholds` (subset of `50,75,100`) | `--alert-thresholds` | ❌ |
| `projectId` | — | ❌ |
| `metadata` (e.g. `environment` for ai-gateway keys) | — | ❌ |

**Follow-up suggestions (again, non-blocking):**

1. **Add `expires_at` and an `ai_gateway_quota` nested attribute** (`limit_amount`, `refresh_period`, `include_byok_in_quota`, `alert_thresholds`). Quotas/budgets are the main reason teams provision AI Gateway keys, so this is the highest-value follow-up. Worth sanity-checking now that adding these later won't be a breaking schema change (nested optional attribute should be fine).

2. **Read back server-side fields.** `GET /v1/api-keys/:id` returns `expiresAt`, `projectId`, and a full `quota` object (`APIKeySchema`/`APIKeyQuotaSchema` in `api/services/api-keys/src/lib/api-key-schema.ts`), but `client.APIKey` here only decodes `id/name/partialKey/teamId/purpose/createdAt` — so no drift detection is possible for anything beyond the name.

3. **Update is more restrictive than the API.** The PR makes `name` `RequiresReplace` and errors in `Update`, but the API supports in-place rename (`update-api-key.ts`) and in-place quota changes (`update-api-key-quota.ts`: `limitAmount`, `refreshPeriod`, `includeByokInQuota`, `alertThresholds`, `active`, `archived`). Forcing replacement on rename rotates the secret `api_key_string`, which is a footgun for keys wired into CI. Rename should become an in-place update, and quota changes should hit the quota endpoint rather than replace the key.

4. **Minor:** `purpose` validated as `OneOf("ai-gateway")` — fine for now, but the API accepts any string (gated by ACL), so new purposes will require a provider release.

Key files for reference:
- CLI: `vercel/packages/cli/src/commands/ai-gateway/api-keys-create.ts`, `util/ai-gateway/{api-keys,quota,expiry}.ts`
- API: `api/services/api-keys/src/endpoints/{create-api-key,get-api-key,update-api-key,update-api-key-quota}.ts`, `src/lib/{create-api-key,api-key-schema}.ts`, `src/lib/ai-gateway/create-quota.ts`
- AI Gateway: `ai-gateway/app/v1/quotas/route.ts`

shaper:
> aiGatewayQuota.includeByokInQuota

side note -- this feature is not yet launched and can be omitted/ignored

client/api_key.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+)
+
+type APIKey struct {
+	ID           string  `json:"id"`
+	Name         string  `json:"name"`
+	PartialKey   string  `json:"partialKey"`
+	TeamID       string  `json:"teamId"`
+	Purpose      string  `json:"purpose"`
+	CreatedAt    int64   `json:"createdAt"`
```
dglsparsons: omit this, maybe.

vercel/resource_api_key.go
```diff
@@ -0,0 +1,263 @@
...
+			"partial_key": schema.StringAttribute{
+				Description:   "The final characters of the API key, used for identification.",
+				Computed:      true,
+				PlanModifiers: []planmodifier.String{stringplanmodifier.UseNonNullStateForUnknown()},
+			},
+			"created_at": schema.Int64Attribute{
+				Description:   "The Unix timestamp in milliseconds when the API key was created.",
+				Computed:      true,
+				PlanModifiers: []planmodifier.Int64{int64planmodifier.UseNonNullStateForUnknown()},
+			},
```
dglsparsons:
```suggestion
```

vercel/resource_api_key.go
```diff
@@ -0,0 +1,263 @@
...
+				},
+			},
+			"purpose": schema.StringAttribute{
+				Description:   "The purpose of the API key. Currently, only `ai-gateway` is supported.",
+				Required:      true,
+				PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplace()},
+				Validators: []validator.String{
+					stringvalidator.OneOf("ai-gateway"),
+				},
+			},
```
dglsparsons: Should we have a specific resource for 
`vercel_ai_gateway_api_key`? 

Are there plans to expand this into other purposes? 

Curious mostly from discoverability here. Open to discussion tho.

client/api_key.go
TooTallNate: Removed in 8fadb4e — the client struct now only carries `id`, `name`, `partialKey`, and `teamId`.

vercel/resource_api_key.go
TooTallNate: Removed the `created_at` attribute in 8fadb4e.

vercel/resource_api_key.go
TooTallNate: Renamed to `vercel_ai_gateway_api_key` in 8fadb4e — the `purpose` attribute is gone and `ai-gateway` is implicit, which matches @shaper's suggestion as well. The API also supports a `sandbox` purpose (gated by ACL), so if other purposes need Terraform support later they can get their own focused resources (e.g. `vercel_sandbox_api_key`) or a generic resource without conflicting with this one.

shaper (approved):
Re-reviewed after the rework — this addresses all of the earlier feedback: dedicated `vercel_ai_gateway_api_key` resource with `purpose` implicit, quota (limit / refresh period / alert thresholds), `expires_at`, `project_id`, in-place rename preserving the secret, in-place quota upsert with archive-on-removal, and quota drift detection via the list-endpoint readback. `includeByokInQuota` correctly omitted. Nice work.

I verified the client behavior against the `vercel/api` implementation (`services/api-keys`): the `until` cursor pagination, quota decoration semantics (archived quotas are filtered out, so archive convergence in `waitForQuota` works), atomic create-with-quota (the API hard-deletes the key if quota creation fails, so Terraform never ends up with a half-created resource), and the rename PATCH body carrying only `name` all check out.

A few **non-blocking** observations, all fine as follow-ups:

1. **Transient quota-read failures show phantom drift.** The list endpoint's quota decoration is best-effort — server-side, quota fetch errors are swallowed and keys are returned *without* quota. `GetAIGatewayAPIKeyQuota` can't distinguish "no quota" from "decoration failed", so a quota-service blip during refresh makes the quota look removed. Harmless (the next apply re-upserts idempotently) but noisy. The real fix is API-side: decorate the get-by-ID endpoint with quota so the provider doesn't need the list scan at all — happy to take that as a follow-up on our side.

2. **`waitForQuota` compares against plan-derived values rather than the PATCH/create response.** If the server ever normalizes a value (rounding, clamping), every apply would burn the full 90s and emit the propagation warning. Comparing against what the API actually returned would converge immediately in that case.

3. **Quota `active` isn't modeled.** A quota paused out-of-band (dashboard/CLI) still decorates with `active: false`; the provider sees no drift and the config implies the budget is enforced. Reasonable to add alongside `includeByokInQuota` once that launches.

Approving — none of the above blocks merge.

dglsparsons (approved):

PR #567, opened 2026-07-29
fix: remove plan-time sensitive env var policy validation

brookemosby:
## Summary
- Removes `ModifyPlan` validation that blocked non-sensitive env vars when a team has the legacy Sensitive Environment Variables Policy enabled.
- Removes plan-time checks that required `sensitive = false` for Development-only targets and blocked sensitive vars on Development.
- Deletes the `environment_variable_sensitivity` helper; sensitivity rules are enforced by the Vercel API at apply time instead of during `terraform plan`.
- Updates resource schema notes/descriptions to reflect API-side enforcement.

Affected resources:
- `vercel_project_environment_variable`
- `vercel_project_environment_variables`
- `vercel_project` (inline `environment`)
- `vercel_shared_environment_variable`

## Test plan
- [x] `go test` for updated unit tests (`resource_project_environment_variable_unit_test.go`, `resource_shared_environment_variable_unit_test.go`, `resource_project_environment_defaults_unit_test.go`)
- [ ] `terraform plan` against a team with sensitive env var policy enabled and `sensitive = false` on a new env var — should no longer fail at plan time
- [ ] `terraform apply` still surfaces API validation errors when team rules reject the configuration

Made with [Cursor](https://cursor.com)

PR #592, opened 2026-08-24
feat: env visibility, disjunctive secret policy, and remove plan-time env validation

brookemosby:
## Summary
- Removes plan-time team sensitive-environment-variable **policy** validation (API lookup). Team policy is enforced by the Vercel API at apply time.
- Keeps the local Development-target check: env vars targeting `development` must set `sensitive = false`.
- Adds `disjunctive_production_secret_policy` (`on` / `off` / `default`) on `vercel_team_config` and its data source.
- Deprecates `sensitive_environment_variable_policy` (still supported). It is **not** replaced by `disjunctive_production_secret_policy` — those enforce different rules.
- Adds optional `visibility` (`config` | `secret`) on project env var resources and inline `environment`. Existing `sensitive`-only configs keep working; when omitted, visibility is not sent to the API.

Supersedes #567.

## Test plan
- [x] `go build ./...`
- [x] Unit tests for visibility mapping and Development `ModifyPlan` rejection
- [ ] `terraform plan` with team sensitive policy enabled + `sensitive = false` on preview/production — should no longer fail at plan
- [ ] `terraform plan` with Development target + `sensitive = true` — should still fail at plan
- [ ] `terraform apply` still surfaces API validation errors when team policy rejects the config
- [ ] Existing configs without `visibility` continue to plan/apply unchanged
- [ ] Setting `disjunctive_production_secret_policy` updates team config; legacy `sensitive_environment_variable_policy` still works


vercel/resource_project_environment_variable.go
```diff
@@ -231,71 +232,6 @@ func (e ProjectEnvironmentVariable) hasTarget(ctx context.Context, target string
 	return false, nil
 }
 
-func (r *projectEnvironmentVariableResource) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
```
dglsparsons, line 234: [Doug's Clanker] The API accepted target=[development] with sensitive=true in all four affected acceptance tests, so removing this validation does not move the rejection to apply; it lets the invalid variable be created. Keep the local Development-target check while removing only the team-policy API lookup, or otherwise ensure the API rejects it.

vercel/resource_team_config.go
```diff
@@ -107,12 +107,22 @@ func (r *teamConfigResource) Schema(_ context.Context, req resource.SchemaReques
 				Description:   "A description of the team.",
 			},
 			"sensitive_environment_variable_policy": schema.StringAttribute{
-				Description:   "Ensures that all environment variables created by members of this team will be created as Sensitive Environment Variables which can only be decrypted by Vercel's deployment system.: one of on, off or default.",
+				Description:        "Ensures that all environment variables created by members of this team will be created as Sensitive Environment Variables which can only be decrypted by Vercel's deployment system. One of `on`, `off`, or `default`.",
+				DeprecationMessage: "This attribute is deprecated. Use `disjunctive_production_secret_policy` instead.",
```
dglsparsons: [Doug's Clanker] These settings enforce different rules: the legacy field forces new environment variables to be sensitive, while the new field separates Production secrets from non-Production targets. Telling users to replace one with the other changes their security policy; deprecate it without this replacement guidance unless there is an actual equivalent successor.

vercel/resource_project_environment_variable.go
brookemosby: Addressed: restored the local Development-target check (`sensitive` must be `false` when `development` is targeted) across the four env var resources, and kept only the removal of the team-policy API lookup. Added a unit test covering the Development plan rejection.

vercel/resource_team_config.go
brookemosby: Addressed: deprecation message no longer tells users to replace this with `disjunctive_production_secret_policy`. It now notes the attribute is deprecated and that the two policies enforce different rules.

vercel/env_var_visibility.go
```diff
@@ -0,0 +1,70 @@
...
+
+func envVarVisibilityFromResponse(envType string, apiVisibility *string) types.String {
+	if apiVisibility != nil && *apiVisibility != "" {
+		return types.StringValue(*apiVisibility)
+	}
+	if envType == "sensitive" {
+		return types.StringValue("secret")
+	}
+	return types.StringValue("config")
+}
```
dglsparsons, line 66: [Doug's Clanker] Legacy plain/encrypted variables can have no visibility; Vercel intentionally infers only sensitive → secret because storage type alone does not imply config. This returns config for an absent API field, so refresh or import records a false classification and a later update can send it back as an explicit config. Return null when visibility is absent and type is not sensitive.

dglsparsons (approved):

PR #593, opened 2026-09-02
Preserve firewall configuration during import

jerryhjones:
`vercel_firewall_config` uses prior Terraform state while converting API responses so optional fields retain their configured shape. `ImportState` has no prior state, so it currently passes an almost-empty reference into that conversion. As a result, fresh imports can omit API-returned values including `enabled`, active custom-rule shape, condition keys, Bot Protection, and non-default OWASP/CRS configuration.

Give the existing API-to-state conversion two explicit modes: preserve configured shape for ordinary create, read, and update operations, and produce canonical API-derived state for `ImportState`. This keeps a single flattening implementation rather than introducing a second API-to-state mapper.

Canonical import mode:

- preserves API `enabled` and custom-rule active values;
- retains non-empty optional descriptions, action durations, condition keys, and keyed scalar values while leaving empty optional fields null;
- preserves Bot Protection and AI Bots without materializing the deprecated `bot_filter` shape;
- retains recognized, non-default OWASP/CRS categories from partial or complete responses, including when CRS is returned without another managed ruleset;
- honors explicit active and inactive OWASP markers while ignoring unknown CRS and managed-rule keys;
- leaves marker-absent, all-default CRS omitted so a round trip cannot accidentally enable OWASP; and
- leaves empty IP-rule notes null.

Ordinary conversion remains in configured-shape mode, preserving the provider's existing null and deprecated `bot_filter` behavior.

## Testing

- `go test ./... -skip '^TestAcc_'`
- `gofmt -s -l .`
- `go vet ./...`
- `go run honnef.co/go/tools/cmd/staticcheck@2026.1 -tags it ./...`
- `go run github.com/bflad/tfproviderlint/cmd/tfproviderlint@v0.31.0 -R018=false ./...`
- `go build ./...`

Read-only import validation against an existing firewall configuration, with no apply:

- generated import plan: `1 to import, 0 to add, 0 to change, 0 to destroy`;
- direct import completed successfully;
- the patched provider produced a zero-change plan;
- two subsequent plans with the unmodified released v5.15.0 provider produced zero changes; and
- imported state matched 22 ordered custom rules, 18 condition keys, firewall enabled state, Bot Protection, and effective semantics for all 11 returned CRS categories.

Focused tests also cover partial/default/unknown CRS responses, active and inactive OWASP markers, scalar/list/existence conditions, empty optional values, IP-rule notes, unknown managed rules, AI Bots, and ordinary configured-shape behavior.

Acceptance tests were not run because they create and delete Vercel resources.

### DIFF
--- a/vercel/resource_firewall_config.go
+++ b/vercel/resource_firewall_config.go
@@ -679,7 +679,14 @@ func (r *FirewallRule) Mitigate() (client.Mitigate, error) {
 	return mit, nil
 }
 
-func fromFirewallRule(rule client.FirewallRule, ref FirewallRule) (FirewallRule, error) {
+type firewallFlattenMode uint8
+
+const (
+	preserveConfiguredShape firewallFlattenMode = iota
+	canonicalImportShape
+)
+
+func fromFirewallRule(rule client.FirewallRule, ref FirewallRule, mode firewallFlattenMode) (FirewallRule, error) {
 	var err error
 	r := FirewallRule{
 		ID:          types.StringValue(rule.ID),
@@ -687,11 +694,11 @@ func fromFirewallRule(rule client.FirewallRule, ref FirewallRule) (FirewallRule,
 		Description: types.StringValue(rule.Description),
 		Active:      types.BoolValue(rule.Active),
 	}
-	if rule.Active && ref.Active == types.BoolNull() {
+	if mode == preserveConfiguredShape && rule.Active && ref.Active == types.BoolNull() {
 		r.Active = ref.Active
 	}
 
-	r.Action, err = fromMitigate(rule.Action.Mitigate, ref.Action)
+	r.Action, err = fromMitigate(rule.Action.Mitigate, ref.Action, mode)
 	if err != nil {
 		return r, err
 	}
@@ -703,7 +710,7 @@ func fromFirewallRule(rule client.FirewallRule, ref FirewallRule) (FirewallRule,
 			if len(ref.ConditionGroup) > j && len(ref.ConditionGroup[j].Conditions) > k {
 				cond = ref.ConditionGroup[j].Conditions[k]
 			}
-			conditions[k], err = fromCondition(condition, cond)
+			conditions[k], err = fromCondition(condition, cond, mode)
 			if err != nil {
 				return r, err
 			}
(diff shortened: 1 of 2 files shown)


vercel/resource_firewall_config.go
```diff
@@ -874,69 +885,76 @@ type IPRule struct {
...
-		RCE:  fromCoreRuleset(conf["rce"], ref.RCE),
-		SD:   fromCoreRuleset(conf["sd"], ref.SD),
-		MA:   fromCoreRuleset(conf["ma"], ref.MA),
-		PHP:  fromCoreRuleset(conf["php"], ref.PHP),
-		GEN:  fromCoreRuleset(conf["gen"], ref.GEN),
-		JAVA: fromCoreRuleset(conf["java"], ref.JAVA),
+	rules := &CRSRule{}
+	set := func(key string, ref *CRSRuleConfig, target **CRSRuleConfig) {
+		crsRule, ok := conf[key]
+		if mode == canonicalImportShape && !ok {
```
dglsparsons, line 899: [Doug's Clanker] After importing a partial CRS response, the next normal Read treats every absent category as a zero-valued rule and materializes it with active=false and an empty required action. The state then changes on refresh and can produce persistent diffs. Check key presence in preserve mode too, leaving an absent category nil when its reference is nil.

vercel/resource_firewall_config.go
```diff
@@ -954,47 +972,48 @@ func fromClient(conf client.FirewallConfig, state FirewallConfig) (FirewallConfi
...
+			if owasp.Active && managedRulesets.OWASP == nil {
+				managedRulesets.OWASP = &CRSRule{}
+			} else if !owasp.Active {
+				managedRulesets.OWASP = nil
+			}
+		}
+		if managedRulesets.BotProtection != nil || managedRulesets.AiBots != nil || managedRulesets.OWASP != nil {
+			cfg.ManagedRulesets = managedRulesets
+		}
+	} else if len(conf.ManagedRulesets) > 0 {
```
dglsparsons, line 1001: [Doug's Clanker] Canonical OWASP state is lost on the next normal Read: a CRS-only response skips this branch because managedRulesets is empty, while an active marker without CRS enters it but never assigns OWASP. Refresh then drops the imported OWASP block and can cause drift or reconfiguration. Reconstruct OWASP from either the marker or CRS instead of requiring both.

dglsparsons:
Fixed in #594. Releasing in v5.15.1. Thanks for raising this!

PR #595, opened 2026-09-08
Add vercel_alert_rule resource

ergunsh:
## Summary

Adds the `vercel_alert_rule` resource, allowing teams to manage built-in Vercel alert rules with Terraform.

Built-in alert rules determine:

- Which anomaly types should raise alerts
- Which projects the rule applies to
- The minimum severity to match
- Whether team owners receive notifications
- The optional incident.io routing key included in configured incident.io webhook payloads

The resource supports create, read, update, delete, and import, along with generated documentation and acceptance-test coverage.

```hcl
  name = "example-project"
}

resource "vercel_alert_rule" "checkout_errors" {
  type = "built-in"
  name = "Checkout 5xx anomalies"

  rule_scope = {
    type        = "include"
    project_ids = [vercel_project.example.id]
  }

  triggers = [{
    type   = "error_anomaly"
    filter = "statusGroup:5xx AND route:/api/checkout"
  }]

  match_minimum_severity_level = "high"

  notification_settings = {
    enable_team_owner_notifications = true
  }
}
```

## Supported configuration

Rules can target:

- Every project in a team with an `all` scope
- A specific set of projects with an `include` scope
- Every project except a specific set with an `exclude` scope

Each rule configures one or more built-in anomaly triggers. Filters are supported for trigger types that accept KQL filtering, including error and usage anomalies.

Custom alert rules are intentionally outside the scope of this change and will be introduced separately once we're launching them widely.

### Stable KQL filters

The Alerts API may return a semantically equivalent KQL filter using different text. For example:

```text
Configured: NOT statusGroup:4xx
Returned:   statusGroup:5xx
```

Since triggers are represented as a Terraform set, replacing the configured text with the returned representation would change the element hash and could produce an inconsistent result or persistent diff.

The resource therefore retains the user-authored filter in Terraform state while recording the API representation in resource private state. This keeps plans stable while still detecting filters changed outside Terraform.

### Compatibility with existing rules

Existing rules can contain trigger modes or filters that remain readable but can no longer be written using the current API policy.

The resource preserves these states during import and refresh without fabricating a new trigger configuration. If triggers are omitted from configuration after import, the existing trigger mode is retained.

Updates are sent as partial PATCH requests containing only fields that actually changed. This allows metadata such as the rule name or notification settings to be updated without resubmitting and potentially invalidating a legacy filter.

### Predictable plans

Stable computed values are retained during updates so they do not appear as unrelated `(known after apply)` changes. This includes:

- `created_at`
- `is_default`
- `trigger_mode`
- Unchanged notification settings

`updated_at` remains unknown until apply because the API generates a new timestamp.

### Notification settings

Notification settings can be omitted during creation, allowing the API to select its defaults. The resolved settings are then stored in Terraform state.

The resource can enable notifications for team owners and configure an optional incident.io routing key. When an incident.io webhook is configured separately, this routing key is included in its webhook payload. Setting the routing key does not create a webhook or control webhook delivery.

Notification channel and webhook configuration remain outside this resource.

### Import

Rules can be imported using a rule ID when the provider has a default team:

```sh
terraform import vercel_alert_rule.example ar_xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
```

Without a provider-level default team, include the team ID:

```sh
terraform import vercel_alert_rule.example team_xxxxxxxxxxxxxxxxxxxxxxxx/ar_xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
```

Alert rules require team context, so personal-account imports are not supported.


vercel/resource_alert_rule.go
```diff
@@ -0,0 +1,779 @@
...
+		// Every writable trigger configuration uses selected mode. Derive this
+		// instead of blindly retaining state: an imported legacy `all` rule must
+		// plan a mode change when the user starts managing its trigger set.
+		resp.Diagnostics.Append(resp.Plan.SetAttribute(ctx, path.Root("trigger_mode"), types.StringValue("selected"))...)
+	}
+}
+
+func (r *alertRuleResource) ValidateConfig(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
+	var config AlertRule
+	resp.Diagnostics.Append(req.Config.Get(ctx, &config)...)
```
dglsparsons, line 590: [Doug's Clanker] If `rule_scope` is an object that is unknown during validation or planning (for example, a module input), `Config.Get` fails with a provider `Value Conversion Error` because `*AlertRuleScope` cannot represent unknown objects. This blocks otherwise valid configurations before the scope checks can defer validation. Store the scope as `types.Object` and decode it only once known, including in `ModifyPlan`.

dglsparsons (approved):

correttojs (approved):

vercel/resource_alert_rule.go
ergunsh: Done!

PR #601, opened 2026-09-14
Add vercel_alert_rule_notification resource

ergunsh:
tl;dr: Adds a dedicated Terraform relationship resource for linking existing Slack channels and account webhooks to Vercel alert rules, including the team default rule. Teams with one Slack installation can omit its ID because the API returns the resolved identity for Terraform state.

## Problem

The `vercel_alert_rule` resource can configure rule-owned notification settings, but users cannot currently manage the reusable Slack channel and webhook destinations linked through the Alerts v3 notification APIs. Managing these links outside Terraform leaves notification routing outside the desired infrastructure state.

Terraform also needs the canonical Slack installation/channel pair after creation. Requiring users to provide the installation ID avoids ambiguity, but is unnecessary when the team has exactly one installation and the API can return the installation it selected.

## Solution

Add `vercel_alert_rule_notification`, which manages one link between an alert rule and either an existing account webhook or a Slack channel. The resource supports create, read, delete, drift detection, and import. It also supports `ar_default`, which cannot be managed by `vercel_alert_rule` itself.

`slack_installation_id` is optional and computed. When omitted, the API selects the team's sole completed Slack installation and the provider stores the returned ID in state. Teams with multiple Slack installations must continue to select one explicitly. Webhooks remain separate `vercel_webhook` resources and must be account-owned subscriptions to `alerts.triggered`.

<details>
<summary>Technical details</summary>

- Adds Alerts v3 client methods for per-rule notification listing and idempotent link/unlink mutations.
- Uses the mutation response's canonical Slack or webhook identity to construct Terraform state without a follow-up read.
- Models webhook identity by `webhookId` and Slack identity by resolved `configId` plus `channelId`.
- Marks `slack_installation_id` as optional and computed, validates the mutation response, and persists the resolved installation before constructing the Terraform resource ID.
- Uses replacement-only destination attributes and validates exactly one webhook or Slack channel per resource.
- Removes state when the rule or destination link is deleted out of band while preserving state on transient API failures.
- Supports provider-team and explicit-team import formats for webhook and Slack links.
- Registers the resource, adds generated documentation and examples, and updates `vercel_alert_rule` guidance to point to it.
- Adds unit coverage for client requests and resolved mutation responses, schema and target validation, CRUD state behavior, drift, and import, plus webhook acceptance coverage.

</details>


examples/resources/vercel_alert_rule_notification/resource.tf
```diff
@@ -0,0 +1,14 @@
+resource "vercel_alert_rule_notification" "slack" {
+  alert_rule_id    = "ar_xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
```
dglsparsons: We should probably show `resource vercel_alert_rule` being created, and then the ID passed into this field.

PR #603, opened 2026-09-21
fix(rolling-release): read native stage rules

windsornguyen:
Creating or refreshing a rolling release policy can preserve the planned stages even when Vercel returns different rules or no policy. Manual writes also send approval flags that the configure request does not use. This change serializes the request fields used by the CLI and reconstructs Terraform state from the native GET response.

After an acknowledged write, the resource reads the policy before recording state. A confirmed null or 404 removes the resource. Missing, contradictory or unsupported rules produce diagnostics instead of copying old state. A lost write reply returns an error without issuing another request.

The provider supports two to ten stages, including a non-final stage whose rules identify manual or automatic advancement. The two-stage minimum is a provider restriction, not a claim about server validation. Existing single-final-stage configurations require a reviewed conversion before upgrading. The generated resource and data-source documentation describes that change.

## Verification

27 focused cases passed with Go race instrumentation against the actual Terraform resource methods and a local HTTP server. The new single-final-stage Read case also passed. Tests cover request bodies, changed native rules, confirmed absence, malformed and contradictory responses, unrepresentable modes, and a real connection close after a write. Both affected packages passed vet. The documentation was regenerated with tfplugindocs v0.25.0 and Terraform 1.14.9.

The original provider failed the preserved request-body, changed-state and absent-state tests. No live provider acceptance test or deployment was performed. Existing HTTP 429 retries are unchanged.

Base: [v5.16.0](https://github.com/vercel/terraform-provider-vercel/releases/tag/v5.16.0). Scope: 490 authored changed lines and 16 generated documentation lines. No dependency changes.


client/project_rolling_release.go
```diff
@@ -66,19 +62,21 @@ func (c *Client) CreateRollingRelease(ctx context.Context, request CreateRolling
...
 	if err != nil {
 		return RollingReleaseInfo{}, fmt.Errorf("error enabling rolling release: %w", err)
 	}
 
-	result.ProjectID = request.ProjectID
-	result.TeamID = request.TeamID
+	// Record the native GET result, not the submitted settings or PATCH acknowledgment.
+	result, err := c.GetRollingRelease(ctx, request.ProjectID, request.TeamID)
+	if err != nil {
+		return RollingReleaseInfo{}, err
```
dglsparsons, line 78: [Doug's Clanker] If the PATCH succeeds but this GET fails (for example, with a transient 503), Create returns before saving any resource identity, leaving the configured policy outside Terraform state. The next apply then sees the existing policy and refuses to proceed without an import. Preserve the acknowledged creation and save enough identity in the resource's error path for Terraform to recover, while still reporting the read failure.

PR #609, opened 2026-09-23
Validate ignore_command length in the schema

robbertstevens:
`ignore_command` has a documented 256-character limit, but the provider only
finds out when the API refuses the update:

```
Could not update project team_xxx prj_xxx, unexpected error:
invalid_command_for_ignoring_build_step - If defined, the command for ignoring
the Build Step must have a maximum length of 256 characters.
```

Since the limit is fixed and documented, the schema can carry it and `plan` can
fail instead:

```
Attribute ignore_command string length must be at most 256, got: 306
```

which names the attribute and the actual length.

### Why it's worth catching early

An ignore command grows by editing. Ours sat at 210 characters for months;
adding one guard took it to 306, and the first sign of trouble was a failed
apply in CI with the project update half-applied. A plan-time error would have
caught it on the machine where the edit was made.

### Notes

- Follows the existing use of `stringvalidator` in `resource_project.go`, and
  `LengthAtMost` as used in `resource_firewall_config.go` and
  `resource_bulk_redirects.go`.
- Added an `ExpectError` step beside the existing `serverless_function_region`
  and `framework` validator cases.
- The data source is untouched — it reads the value rather than sending it.
- I verified the validator against the resource schema directly: 256 produces
  no diagnostic, 257 and 306 produce one each. I could not run the acceptance
  tests, which need `VERCEL_API_TOKEN`; `go build`, `go vet` and `gofmt` are
  clean.


dglsparsons (approved):

dglsparsons:
Superseded by #610, which recreates this change with a signed commit. Thank you for the contribution.

